PR #1, opened 2016-04-26
Checking in ... everything.

nameofname:


nameofname:
Sadly, I cannot merge this.

README.md
```diff
@@ -1,2 +1,70 @@
...
+    	loaders: [
+        	{ test: /\.(js|jsx|svg)$/, loader: 'babel' },
+        	{ test: /\.svg$/, loader: 'svg-react' }
+    	],
+    	resolve: {
+        	alias: {
+            	react-dom: __dirname + '/node_modules/react'
+	        }
+    	}
+	}
```
hellatan: it seems that some of your code blocks have some weird alignment issues.

js/SvgWrapper.es.js
```diff
@@ -0,0 +1,45 @@
...
+    template : _.template(tmpString),
+
+    render : function () {
+        this.$el.html(this.template(this.templateVars()));
+        return this;
+    },
+
+    initialize : function (options) {
+        console.assert(options.icon, "Icon is required to use SvgWrapper BB component.");
+        this.options = options;
```
hellatan: attach the properties you want to the instance instead of blindly trusting that all the options will be there.

js/SvgWrapper.es.js
```diff
@@ -0,0 +1,45 @@
...
+ * new SvgWrapper({
+ *      className : 'icon-100',
+ *      icon : require('!html!dibs-static/1stdibs/fonts/svg-icons/src/logo-no-reg.svg')
+ * });
+ *
+ * @type {void|*|Object}
+ */
+const SvgWrapper = Backbone.BaseView.extend({
+
+    template : _.template(tmpString),
```
hellatan: why not just lodash.template - https://lodash.com/docs#template

js/SvgComponent.jsx
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const React = require('react');
+const { PropTypes } = React;
+
+const SvgComponent = props => {
+    let className = 'icon ' + props.className + '-icon ' + props.className;
```
hellatan: the `icon` class name is too generic. `dibs-vg`?

js/SvgWrapper.es.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+const _ = require('underscore');
+const Backbone = require('dibs-backbone');
```
nameofname: @robrichard @hellatan @twhid - not sure what to do about this, I don't want to include dibs-backbone in our public repo, since it's not supported any more, but I also don't want to require backbone here since it would get added twice to our bundles. Any ideas on how to get around that? Alias backbone to dibs-backbone in v2?

js/SvgWrapper.es.js
```diff
@@ -0,0 +1,45 @@
...
+    },
+
+    initialize : function (options) {
+        console.assert(options.icon, "Icon is required to use SvgWrapper BB component.");
+        this.options = options;
+        return this;
+    },
+
+    templateVars : function () {
+        return Object.assign({
```
hellatan: You're not merging any objects together. You can return just the object itself.

js/SvgWrapper.es.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+const _ = require('underscore');
+const Backbone = require('dibs-backbone');
+const tmpString = "<span class='icon <%= className %> <%= className %>-icon <%= size ? 'icon-'+size : '' %>' title='<%= title %>'><%= icon %></span>";
```
hellatan: what are some possible outputs for `'icon-' + size` ?

scss/icons.scss
```diff
@@ -0,0 +1,24 @@
...
+    height: 100px;
+}
+.icon.icon-50 {
+    height: 50px;
+}
+.icon.icon-25 {
+    height: 25px;
+}
+.icon.icon-16 {
+    height: 25px;
```
hellatan: i feel like these sizes are contrived and don't add much to the overall module.

scss/icons.scss
```diff
@@ -0,0 +1,24 @@
+// Styling for icons including basic colors and common sizing.
+
+.svg-icon {
```
hellatan: i think this may conflict with existing classes we are using internally. Maybe call this one `dibs-vg-icon`?

src/bullet-point.svg
```diff
@@ -0,0 +1,7 @@
+<svg version="1.1" class="svg-icon" id="bullet-point" viewBox="0 0 250 250" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <g>
+        <path fill="#F2EAD5" d="M125,249.3C56.4,249.3,0.7,193.6,0.7,125S56.4,0.7,125,0.7S249.3,56.4,249.3,125S193.6,249.3,125,249.3z
```
hellatan: ugh..i hate this. there needs to be something better but this can probably wait for another day...

README.md
```diff
@@ -1,2 +1,70 @@
...
+	// ...
+	<IconWrapper className="icon-spin" size={25} title={"Weeee! Weeeeeeeee!"}>
+        <Icon />
+    </IconWrapper>
+
+
+
+## In Backbone : 
+
+If you are using Webpack and want to render your SVG with Backbone we recomment using the [html loader](https://www.npmjs.com/package/html-loader) which returns the SVG file as a string, or the [webpack compile templates loader](https://www.npmjs.com/package/webpack-compile-templates) which will transform the SVG file to an underscore template.
```
twhid: typo: recommend

README.md
```diff
@@ -1,2 +1,70 @@
...
+    	}
+	}
+
+One gotcha I ran into while implementing is that svg-react-loader requires that you also use babel loader, so this line : 
+
+	{ test: /\.svg$/, loader: 'babel!svg-react' }
+
+will not work if you already have babel configured as a loader in your config. To solve, add ".svg" to the list of extensions to match against, and remove it from that line to end up with something like : 
+
+	module.exports = {
```
twhid: Does anyone else hate how these loaders work in webpack? Why not simply require the packages instead of their own arcane syntax: `babel!svg-react` ?

js/SvgWrapper.es.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * A very simple wrapper for an SVG icon required from the static repository.
+ * NOTE! In v2 the only way to use icons in back-bone ie. without react is to require using a different loader
+ * (for webpack) - you *must also use the leaging "!" in your loader so the default SVG->React loader is not used.
+ *      Example :
+ *
+ * new SvgWrapper({
+ *      className : 'icon-100',
+ *      icon : require('!html!dibs-static/1stdibs/fonts/svg-icons/src/logo-no-reg.svg')
```
twhid: Using this webpack `!html` construct going to make server rendering impossible/hard ?

twhid: If so, we may need a wrapper *around* the wrapper -- one for server and one for client..

hellatan: re: server-side rendering - i think you're right, @twhid. There would need to be some hook for this that node can understand. This is the problem we are running into when requiring scss files on the server.

js/SvgWrapper.es.js
hellatan: if you're trying to not make it mutable after returning, then use `Object.freeze` instead

js/SvgWrapper.es.js
twhid: We'll probably never render Backbone components on the server so this isn't the biggest issue. But we need to keep this in mind as we develop React components that we want to render on the server.

js/SvgWrapper.es.js
twhid: You have a bigger problem than that, you're also extending `BaseView` -- if you want to make this portable, you should extend `Backbone.View`.

To get around requiring our Backbone shim, you could have `Backbone` passed to this module when it's required (a la angular dependency injection).

.gitignore
```diff
@@ -0,0 +1,13 @@
+.project
```
twhid: This list is random.

README.md
nameofname: Just surfacing it as an option to the reader. I don't like it but that's why I implemented in the webpack config instead.

js/SvgWrapper.es.js
nameofname: @twhid agree - that's why I was posing the question. Honestly I am not sure that having this BB view here is adding much that we would use. I might just have it require backbone instead of dibs-backbone, then make a module in v2 that we would use instead of this one. If I'm going to do that is there any real value to having this here? Should I just delete it?

twhid: It's a nice example of how you would use it, maybe just move it to the documentation or an example app.

nameofname:
OK - @twhid @hellatan I decided to make a sample file, serves as documentation and a page you can open up to view all the icons. A couple things still need work / cleaning up, but for the most part the sample page is done. It answers the question of what I was doing with the react component, etc. I can move files into different folders and update the readme to make it more clear.

Please take a look, any suggestions are welcome. It's a WIP but feedback is helpful at this point.

.gitignore
nameofname: I updated the list. There is still some stuff in there that came from another gitignore I looked at, but I think it's all useful since it's mostly like .DS_Store / IDE / npm files that get put there by random processes.

sample/compiled/css/icons.css
```diff
@@ -0,0 +1,22 @@
+.dibsvg-svg {
+  max-width: 100%;
+  max-height: 100%;
+  width: auto;
+  height: auto; }
+
+.dibsvg-icon {
+  height: 30px; }
+
+.dibsvg-icon.dibsvg-icon-100 {
```
nameofname: @hellatan is there a way for SCSS to do some magic with the height to make it more dynamic? Would you guys prefer that I just have the SVG wrapper component use the size prop and apply a height style rule directly in the markup?

.gitignore
```diff
@@ -0,0 +1,10 @@
...
+npm-debug.log
+.settings
+.idea
+.idea/*
+/nbproject/private/
+.DS_Store
+node_modules/
+npmtmp
+.sass-cache*
```
hellatan: - is `.sass-cache` even applicable?
- `.idea` should be enough
- what's `/nbproject/private/`?

You don't need the trailing slash for an dir you're ignoring

sample/js/reactSample.jsx
```diff
@@ -0,0 +1,14 @@
+"use strict";
+
+
+const React = require('react');
+const ReactDOM = require('react-dom');
+const IconContainer = require('./IconContainer.jsx');
+
+const render = () => {
+    ReactDOM.render(<IconContainer />, document.getElementById('IconContainer'));
```
hellatan: the `ReactDOM` should be in the consuming file. You really just need this:

```jsx
module.exports = () => {
    return <IconContainer />;
}
```
Then in the file that is require this file, you can do

```js
const ReactSample = require('./reactSample.jsx');
window.onload = function () {
    ReactDOM.render(ReactSample, document.getElementById('IconContainer'));
};
```
I know it's just a sample file, but the sample file will be taken as the source of truth and then this pattern will end up being cargo culted.

sample/compiled/css/icons.css
hellatan: I'm not following how this is being compiled so hard to say. Maybe make the size names more generic like `sm`, `md`, `lg`, etc? One day, size `100` will need to be turned into size `110` and then that breaks the whole naming pattern.

.gitignore
nameofname: Yeah I added sass-cache, the build step is adding it because I'm using sass directly and not webpack. Updated the other 2.

sample/compiled/css/icons.css
nameofname: @hellatan - yeah, don't disagree. I was more thinking these are useless in general b/c each client will need different sizes. Would it be better to pass the size to the react component and just put it in as a style rule. I kind of think I should do that...

hellatan: @nameofname don't think...just do...😉

js/SvgComponent.jsx
```diff
@@ -0,0 +1,38 @@
...
+ *              </SvgComponent>
+ *
+ *
+ * @param props
+ * @returns {XML}
+ * @constructor
+ */
+const SvgComponent = props => {
+    const compoundClass = 'dibsvg-icon ' + (props.className ? props.className : '');
+    const css = props.size ? {height: props.size, width : props.size} : {};
```
nameofname: @hellatan when you pass size now I apply it to width and height in the svg wrapper span tag. Same thing in the backbone wrapper. No more sizes in the style sheets.

scss/icons.scss
```diff
@@ -0,0 +1,20 @@
+// Styling for icons including basic colors and common sizing.
+
+$grey : #21242b;
```
nameofname: @hellatan added grey and gold colors to the icon style sheets so that other people can be on brand like us. I can remove if you want.

scss/icons.scss
```diff
@@ -0,0 +1,20 @@
...
+$gold : #b19b57;
+
+.dibsvg-icon {
+    display: inline-block;
+    fill: $grey;
+    height: 30px; /* default height, gets over-ridden by style rule in component */
+
+    .dibsvg-svg {
```
hellatan: i don't think you'd need this nested if you're namespacing everything with `.dibsvg-*`

js/SvgWrapper.js
```diff
@@ -0,0 +1,49 @@
...
+    },
+
+    initialize (options={}) {
+        console.assert(options.icon, "Icon is required to use SvgWrapper BB component.");
+        this.options = options;
+    },
+
+    templateVars () {
+        let style = '';
+        if (this.options.size) {
```
nameofname: Backbone wrapper is now doing the same thing with sizing, ie. width and height.

scss/icons.scss
hellatan: maybe call them something different than their names? light, dark?

scss/icons.scss
```diff
@@ -0,0 +1,20 @@
...
+        max-width: 100%;
+        max-height: 100%;
+        width: auto;
+        height: auto;
+    }
+
+}.gold {
```
hellatan: i think you need a new line here

js/SvgWrapper.js
hellatan: es6 these strings, yo!

```js
style = `width: ${this.options.size}px; height: ${this.options.size}px`;
```

js/SvgWrapper.js
```diff
@@ -0,0 +1,49 @@
...
+            style += "width:" + this.options.size;
+            style += "px;height:" + this.options.size + "px";
+        }
+        return {
+            className : this.options.className || '',
+            icon : this.options.icon,
+            title : this.options.title || '',
+            style : style
```
twhid: es6 [shorthand property names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer)...
```js
return {
// ...
    style
}
```

scss/icons.scss
nameofname: Not needed but appropriate? These width and height styles wouldn't do anything outside of dibsvg-icon.

scss/icons.scss
```diff
@@ -0,0 +1,21 @@
+// Styling for icons including basic colors and common sizing.
+
+$icon-color-dark : #21242b;
```
nameofname: @hellatan updated the color names.

sample/js/sample.js
```diff
@@ -0,0 +1,15 @@
...
+const ReactIcons = require('./ReactIcons.jsx');
+const BackboneIcons = require('./BackboneIcons.js');
+require('../../scss/sample.scss');
+
+window.onload = function () {
+    // React :
+    ReactDOM.render(<ReactIcons />, document.getElementsByClassName('sample-container')[0]);
+
+    // Backbone :
+    document.getElementsByClassName('sample-container')[0].appendChild(new BackboneIcons().render().el);
```
nameofname: @hellatan @twhid - I added a BackBone example of how the app would work.

js/SvgWrapper.js
nameofname: Cool, I've never got a chance to use them before...

js/SvgWrapper.js
```diff
@@ -0,0 +1,45 @@
...
+        return this;
+    },
+
+    initialize (options={}) {
+        console.assert(options.icon, "Icon is required to use SvgWrapper BB component.");
+        this.options = options;
+    },
+
+    templateVars () {
+        const style = (this.options.size) ? `width: ${this.options.size}px; height: ${this.options.size}px` : '';
```
nameofname: es6 template string in a ternary yo'

scss/icons.scss
hellatan: i don't think appropriate. The class name is very specific and deliberate. 

```scss
.dibsvg-icon { ... }
.dibsvg-svg { ... }
```
that should suffice instead of needing this unnecessary specificity.

nameofname:
Once we merge, I can use this to create a project page so that users can view the SVG files : 
https://help.github.com/articles/creating-pages-with-the-automatic-generator/

js/SvgWrapper.js
```diff
@@ -0,0 +1,45 @@
...
+        return {
+            className : this.options.className || '',
+            icon : this.options.icon,
+            title : this.options.title || '',
```
hellatan: you can destructure these values if you want:

```js
templateVars() {
    const { className: '', icon: '', title: '', size } = this.options;
    const style = size ? `...` : '';
    return {
        className,
        icon,
        title,
        style
    }
}
```

hellatan: actually use `=` instead of `:`: `{ className = '', ... }`

js/SvgWrapper.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+const template = require('lodash.template');
+const Backbone = require('backbone');
+const tmpString = "<span class='dibsvg-icon <%= className %>' title='<%= title %>' style='<%= style %>'><%= icon %></span>";
```
twhid: Just compile it here, it doesn't change so no need to do it on every instance.
```js
const template = require('lodash.template')(tmpString);

// later
const SvgWrapper = Backbone.View.extend({
    template,
    // etc
}
```

PR #3, opened 2016-05-05
tmpString is a horrible name for a function

twhid:
There doesn't seem to be any way to test the Backbone example...


twhid:
Tested this locally, sample renders properly.

PR #4, opened 2016-05-05
Added sample images to readme

nameofname:
Added some sample images and description to the readme. 

I am also going to create the project page and link to it from the readme so hold off on merging until I do that.


README.md
```diff
@@ -12,6 +14,14 @@ We also provide some basic styling and wrapper modules for React and Backbone.
 
 New icons added to the set will be considered a minor version bump. Changes to existing icons will be a major version.  
 
+## Customize it, style it, resize it : 
+
+Inline SVGs are great, not only because they're more performant, but they're also easier for developers to manipulate. Of course you can scale them infinitely in either direction just by changing some CSS, but having different paths inside of your SVG files also allows you to apply colors or other transformations to individual parts of the SVG. Check out this example : 
+
+![](./readme_imgs/do-not-edit-dark-grey.jpeg)
```
twhid: Why not use the actual SVG here? Seems strange to illustrate your point with bitmap images.

nameofname: There is no way to control the size, they end up huge. 

I could ... I think I know a way to work around it, but I'm not sure I would be able to ... hold on.

twhid: inline CSS won't fix it?

PR #7, opened 2016-05-05
moved sample.html up to the top level so that gh-pages will work

nameofname:
I moved sample.html up to the top level and re-named to index.html so that it will automagically be used in the gh-pages branch as a project page by github.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,7 @@
 <html>
 <head>
     <title>Dibs SVG Sample</title>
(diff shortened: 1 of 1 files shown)


nameofname:
As it turns out, gh-pages will not work, because we have this in place : 
https://github.com/1stdibs/1stdibs.github.io

Still moving this up to the top level, just seems right to me.

PR #8, opened 2016-05-05
Added new props to svg component, etc.

nameofname:
Added width and height to svg wrapper and backbone icon wrapper, as well as option for inline which defaults to true. Implemented in response to learnings from @ascagnel 1st appempt at implementation. 

FYI @hellatan @twhid


nameofname:
@hellatan @twhid also I realized that es6 template strings are so great that I removed lodash.template as dependency.

js/SvgComponent.jsx
```diff
@@ -10,28 +10,36 @@ const { PropTypes } = React;
  *      const Icon = require('dibs-vg/src/clock.svg');
  *      const SvgComponent = require('dibs-vg/js/SvgComponent.svg');
  *      // ... then inside your React component, you can do...
- *              <SvgComponent size={25} className={myClass}>
+ *              <SvgComponent width={10} height={10} className={myClass} inline={false}>
```
hellatan: if the property is a boolean, the convention is to just not include the property if it's set to false. If it's supposed to be true, you just pass the property along.

js/SvgComponent.jsx
```diff
@@ -10,28 +10,36 @@ const { PropTypes } = React;
...
-const SvgComponent = props => {
-    const compoundClass = 'dibsvg-icon ' + (props.className ? props.className : '');
-    const css = props.size ? {height: props.size, width : props.size} : {};
-        return (
-        <span className={compoundClass} title={props.title} style={css}>
-            {props.children}
+const SvgComponent = ({ children, className, width, height, title, inline=true }) => {
+    const compoundClass = `dibsvg-icon ${className}`;
+    const css = {width, height};
+    if (inline) css.display = 'inline-block';
```
hellatan: curly braces:

```js
if (inline) {
...
}
```

js/SvgComponent.jsx
nameofname: It defaults to true so you have to specify false if you want to turn it off. Thought we wanted it to default to true.

ascagnel: Why not name the property after the negative -- pass in a prop called `disableInline` to turn it off?

hellatan: ah ok. Ya, i see what you mean

ascagnel: You could also take a `display` optional prop of type string that defaults to `inline block`, so the if statement below could be replaced with

```
css.display = this.props.display || 'inline-block';
```

js/SvgComponent.jsx
hellatan: @nameofname  see @ascagnel's suggestion above - https://github.com/1stdibs/dibs-vg/pull/8/files/1276e460439920c9f9411629681a77d9689fba7a#r62339050

js/SvgComponent.jsx
nameofname: @ascagnel I like that better. Implementing...

nameofname:
@hellatan @ascagnel ready.

js/SvgComponent.jsx
```diff
@@ -10,28 +10,35 @@ const { PropTypes } = React;
...
         </span>
     );
 };
 
 SvgComponent.propTypes = {
     className: PropTypes.string,
-    size: PropTypes.number,
+    width: PropTypes.number,
+    height: PropTypes.number,
+    inline: PropTypes.bool,
```
hellatan: i don't think you have the linter running here b/c i think it would have caught this issue...i think - need to change this to `display` and type is string

js/SvgComponent.jsx
```diff
@@ -10,28 +10,35 @@ const { PropTypes } = React;
...
  * @returns {XML}
  * @constructor
  */
-const SvgComponent = props => {
-    const compoundClass = 'dibsvg-icon ' + (props.className ? props.className : '');
-    const css = props.size ? {height: props.size, width : props.size} : {};
-        return (
-        <span className={compoundClass} title={props.title} style={css}>
-            {props.children}
+const SvgComponent = ({ children, width, height, title, display='inline-block', className='' }) => {
```
ascagnel: +1

js/SvgComponent.jsx
nameofname: Shiiiit. Thanks for catching that! Fucking fuck, I do not have the linter rocking in this project.

hellatan:
published: 

> daletan:~/projects/dibs-vg (master)
$ npm publish
\+ dibs-vg@0.0.2

PR #9, opened 2016-05-09
Added link to gh-pages in readme.

nameofname:
Just adding a link to the readme. I pushed up the gh-pages branch just to see if it would work, and to my surprise, it did!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ![](https://cdn.rawgit.com/1stdibs/dibs-vg/master/src/logo-no-reg.svg)
 
 # dibs-vg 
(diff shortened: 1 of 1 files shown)


nameofname:
For reference, this is how it's working : 
https://help.github.com/articles/creating-project-pages-manually/

The URL should actually be 1stdibs.github.io/dibs-vg, but 1stdibs.github.io gets redirected to our blog on the basis of the repository by the same name. Any repo with a branch named gh-pages and the correct setup will get such a page, strangely because of the DNS behavior, instead of ending up on 1stdibs.github.io, they end up on beautifulcode.

PR #11, opened 2016-05-10
Doctored icons

nameofname:
I fixed up a whole mess of icons. This includes : 
- squaring up the icon frames
- removing un-necessary attributes from SVG tag
- adding in some new icons I found
- re compiling changed files.


nameofname:
Since I modified a bunch of icons, I think we should at the least minor version bump. I would almost think we should major version bump since we said we would do that if any changes were "breaking" -- meaning we updated the icons themselves.

twhid:
There was at least one class name changed so this should be a major version bump IMHO. Also, offline you indicated that some of the bounding boxes changed shape (rectangular to square) which will affect layouts that were depending on that bounding box.

nameofname:
@twhid that's what I was thinking, good points all, version number bumped.

PR #12, opened 2016-05-13
FINDING-7696 Adding new trade project editor SVGs

ascagnel:
cc @twhid @hellatan 

Adding the new trade project editor SVGs.


src/gear-filled.svg
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
nameofname: clean these guys up to look more like the other SVGs : 
- kill the xml tags and comments
- the svg tag only needs the attributes version, class, viewBox, xmlns, xmlns:xlink, I usually remove the x/y attributes, definitely remove the ID, the style tags and xml:space. Honestly I don't quite know what they all do, but I would rather have these new ones be exactly consistent with the other ones, and if we find that we need those attributes we can add to all at once. 
-- here is an example from the arrow down SVG file : 
```html
<svg version="1.1" class="dibsvg-svg dibsvg-arrow-down" viewBox="0 -8.5 32 32" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```

ascagnel: Should I add a dibs-vg specific class to each file (eg `class="dibsvg-svg dibsvg-gear-filled`)?

nameofname: yes

nameofname:
Looking better, still lacking dibsvg-svg classes.

ascagnel:
Added classes, but I still need the new SVGs with paths.

nameofname:
@ascagnel reading your last comment I am thinking I should wait on merging this? It looks ready to me, just waiting on your word.

ascagnel:
@nameofname wait on merging this.  I haven't received the updated SVGs from the design team yet.

ascagnel:
@nameofname updated with the latest SVGs, everything looks good now
screen shot 2016-05-18 at 8 56 14 am

src/gear-filled.svg
```diff
@@ -0,0 +1,12 @@
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
```
nameofname: Use the class dibsvg-svg then we are ready to go. Example I grabbed from another SVG file : 

```html
```

ascagnel: Fixed.

ascagnel:
Hold off a bit, just got a revision

ascagnel:
Should be good to go now.

nameofname:
@ascagnel merged, versioned, published.

PR #13, opened 2016-05-13
removed all instances of annoying st0 class

nameofname:
cc @hellatan @ascagnel - I removed all of the st0 classes applied by adobe illustrator. Andrew found that these classes were getting applied by adobe and in some cases adobe also adds on style tags that apply styles (ie. fill color) to ALL elements with class of st0 - which is BS. For now developers will have to manually remove all the cruft illustrator adds before committing SVG icons to dibs-vg.


twhid:
Was this entire PR just removing the `st0` class?

nameofname:
yes

nameofname:
@twhid I think it's silly to do a major version bump for this - thoughts?

ascagnel:
this isn't really a breaking change -- nothing should have been writing to the `st0` class on the consuming pages, so it's more fixing potential future bugs.

twhid:
Unless someone was relying on those classes (very doubtful) then this
should be a patch update.

On Fri, May 13, 2016 at 2:02 PM ron williams <notifications@github.com>
wrote:

> @twhid <https://github.com/twhid> I think it's silly to do a major
> version bump for this - thoughts?
>
> —
> You are receiving this because you were mentioned.
>
>
> Reply to this email directly or view it on GitHub
> <https://github.com/1stdibs/dibs-vg/pull/13#issuecomment-219116527>
>
-- 
Tim Whidden | VP, Engineering
1stdibs
tim@1stdibs.com
212.627.3927 (ext. 178)

nameofname:
Ok solid - patch versioned and published @ascagnel @twhid

PR #16, opened 2016-05-23
Pre-compile modules

nameofname:
@twhid I found out when I tried to put my work for SVG into V2 and use the html loader( that would be the require("!html!**"); syntax ) for webpack that it blows up unit tests. 

This is to side-step that. Now instead of trying to get webpack to require the file using a special loader, I just pre-compile all the SVG files into modules that are exported as multi line strings using back ticks.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,4 +6,3 @@ npm-debug.log
 .DS_Store
 node_modules/
 .sass-cache
-dist/*
--- a/generateDist.js
+++ b/generateDist.js
@@ -9,7 +9,8 @@ const svgs = fs.readdirSync(process.cwd() + '/src');
 const dest = process.cwd() + '/dist/templateString';
 
 svgs.forEach(file => {
-    const fileName = dest + '/' + file;
+    const name = file.split('.')[0];
+    const fileName = dest + '/' + name + '.js';
     const leFile = process.cwd() + '/src/' + file;
     const contents = ["module.exports = `", fs.readFileSync(leFile), "`;", "\n"];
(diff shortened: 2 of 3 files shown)


twhid:
Talked to @nameofname -- we're going to fix the issue with tests borking with a require hook instead of duplicating all the svg files.

.npmignore
```diff
@@ -0,0 +1,9 @@
+.project
```
nameofname: @hellatan does this look correct to you? I recall you mentioning to me that .gitignore'd files get ignored by npm. 

I took the advice from isaacs here https://github.com/npm/npm/issues/4778

generateRawModules.js
```diff
@@ -0,0 +1,20 @@
+"use strict";
+
+var fs = require('fs');
+
+/**
+ * Builds the raw_modules directory - each file will be a JS module that uses back ticks to export a multi line string.
+ */
+const build = () => {
```
twhid: Not clear to me why you need to assign this function to a `const` and then immediately invoke it. Why not, you know, not do that?

.npmignore
hellatan: ya, i think this is fine. The `dist

hellatan: ya, i think this is fine. The `dist` folder is what you want, i believe?

nameofname: Yeah

nameofname:
@twhid - changes have been incorporated. I tested by running 
```
npm run prepublish
```
Everything looks like it's working.

PR #19, opened 2016-05-31
GOODS-4335 - adding repost/reject svgs

prestifidibs:


src/reject.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
```
nameofname: remove these XML tags, remove the style tags from the SVG files themselves, also remove the style block. Check out the following example and make sure your SVG tag has the correct attributes ( it should have exactly the same attributes - all SVG files get the same treatment b/c Adobe Illustrator adds all sorts of crap). 

```html
<svg version="1.1" class="dibsvg-svg dibsvg-baby-bottle" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```

Looks like you figured out the class name 👍 . Looks like we don't use xml:space attribute. xmlns attribute has questionable benefit but we are keeping it for now.

prestifidibs:
@nameofname bump

nameofname:
@prestifidibs - I merged then built the sample page. The icons are totally messed up, seems like the viewBox attribute is making the bounding box of both those icons incorrect. Who gave those icons to you?

prestifidibs:
Katy supplied them https://1stdibs.atlassian.net/browse/GOODS-4335

PR #21, opened 2016-06-21
FINDING-8448 Adding empty project icon

ascagnel:


hellatan:
use a hyphen in the name

PR #23, opened 2016-06-29
remove babel-rc add an svg react component that works isomorphically

walfly:
The webpack way does not work on node.  this allows react to expect the svg as a string.

### DIFF
--- a/js/SvgStringToReact.jsx
+++ b/js/SvgStringToReact.jsx
@@ -1,0 +1,29 @@
+const React = require('react');
+const { PropTypes } = React;
+
+const SvgStringToReact = React.createClass({
+    propTypes: {
+        svg: PropTypes.string,
+        styles: PropTypes.object,
+        width: PropTypes.number,
+        height: PropTypes.number
+    },
+    render() {
+        let styles = {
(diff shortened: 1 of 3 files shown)


js/SvgStringToReact.jsx
```diff
@@ -0,0 +1,29 @@
...
+            verticalAlign: "middle"
+        };
+        return (
+            <svg 
+                viewBox="0 0 24 24"
+                width={this.props.width}
+                height={this.props.height || this.props.width}
+                preserveAspectRatio="xMidYMid meet"
+                fit
+                style= {Object.assign({}, styles, this.props.style)}
```
twhid: Should be `this.props.styles` based on the name in `propTypes`.

Probably want tests to test that sort of thing.

twhid:
Put a milestone on this?

js/SvgStringToReact.jsx
walfly: I'll add tests to the repo

walfly:
Closing because not actually needed. https://github.com/1stdibs/dibs-vg/blob/master/generateDist.js

PR #25, opened 2016-07-05
Update README.md

walfly:
I think that people are going to skim this looking for relevant information.  They're going to jump straight to either the backbone or react section which previously would have led them to implement the incorrect webpack way.  I also think versioning and customization should be pushed out of the way because they are less important to everyone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,22 +9,11 @@ Dibs VG are broken up into individual SVG files so you can require them to be in
 
 We also provide some basic styling and wrapper modules for React and Backbone. 
 
-
-## Versioning : 
-
-New icons added to the set will be considered a minor version bump. Changes to existing icons will be a major version.  
-
-## Customize it, style it, resize it : 
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,12 +28,10 @@ const stringSvg = require('dibs-vg/dist/templateString/account-filled.es.js');
...
 Require these files directly in your modules and use just like you would any React component or string for inserting into pages. Pre-compiled React components accept a className prop so you can apply custom styling. 
 
-## Methods for loading with Webpack (not recommended):
+### loading with Webpack (not recommended):
 
 You don't have to use pre-compiled modules. If you want to grab the raw SVG files and manipulate them go ahead, or if you want to use webpack to load the SVG files, use the method(s) below.
 
-### With React / JSX : 
-
 If you are using Webpack, and want to render your SVG with React we recomment [svg-react-loader](https://github.com/jhamlet/svg-react-loader) which will allow you to require your SVG files direclty in your JS as React components.
```
twhid: *recommend (there's a couple other instances of this typo, please fix w this PR)

PR #26, opened 2016-07-06
change the dist react components to be in plain js becasue that is mo…

walfly:
…re expected behavior from a library

Keeping the jsx files so its not breaking

### DIFF
--- a/generateDist.js
+++ b/generateDist.js
@@ -6,6 +6,7 @@
 
 const fs = require('fs');
 const htmlToReact = require('./htmlToReact');
+const babel = require('babel-core');
 const svgs = fs.readdirSync(process.cwd() + '/src');
 const stringDest = `${process.cwd()}/dist/templateString`;
 const reactDest = `${process.cwd()}/dist/react`;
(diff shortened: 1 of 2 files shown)


twhid:
@walfly is going to make this all sync, then we'll merge.

walfly:
Mostly synchronous I didn't want to change the html to react function.

generateDist.js
```diff
@@ -16,29 +17,37 @@ const checkError = err => {
...
 
-    fs.readFile(leFile, (err, data) => {
-        checkError(err);
+    const contents = fs.readFileSync(leFile, {encoding: 'UTF-8'});
 
-        // generate template string module :
-        const fileContents = data.toString();
-        const stringContents = `module.exports = \`${fileContents}\`;`;
+    // generate template string module :
+    const stringContents = `module.exports = \`${contents}\`;`;
```
twhid: Why not use actual quotes inside the template literal (I understand that this isn't your code, but escaping the template literal ticks seems uselessly complex). I can't think there'd be a reason to do it.

twhid: @walfly just that one thing then I'll merge it.

walfly: i think multiline strings is the reason

walfly: but it shouldn't make a difference either way.

twhid: Ah yeah, I think you're right.

walfly: actually i think it breaks the other way.  you'd have have to escape all strings in the html and add new lines.

PR #27, opened 2016-07-08
FINDING-8588 Adding trade photo upload icons

ascagnel:
Relied on by https://github.com/1stdibs/dibs-cms-trade/pull/127


ascagnel:
Published dibs-vg@1.7.0

PR #28, opened 2016-07-08
add a plain js version for template stings too

walfly:


walfly:
I kind of need this, so if there are not objections im going to merge and publish a patch.

PR #30, opened 2016-07-16
removed `dist` from source control

hellatan:
We shouldn't need to track this directory in source control. Since we have the `.npmignore` file, the publish step will still create the `dist` dir for us

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@ npm-debug.log
 .DS_Store
 node_modules/
 .sass-cache
-dist/*
+dist


nameofname:
@hellatan you sure publish will create the dist folder for us? Did you test it out?

walfly:
@nameofname if you have a .npmignore file npm will use that to decide what to publish instead of the gitignore. As long as dist is written before npm publish is run everything will be great.

nameofname:
@hellatan that's cool, I'm concerned that the generateDist.js script will fail if the dist folder is not present at the time of publishing. Can you update that script to make the dist directory for the developer? https://nodejs.org/api/fs.html#fs_fs_mkdirsync_path_mode

nameofname:
@hellatan I was going to merge this but then I realized it breaks the build script if there is no dist folder present. I re-did this PR and added a small script to create the dist folders if none exist : 
https://github.com/1stdibs/dibs-vg/pull/55

hellatan:
sweet. thanks

PR #31, opened 2016-07-17
Updated icon page

hellatan:
- changed some npm scripts
- added icon names
- added icon search (only works in react icons section)
- added hot module replacement and webpack dev server for dev purposes
- css modules (local css)


js/SvgWrapper.js
```diff
@@ -9,8 +9,8 @@ const template = function ({className, icon, title, style}) {
 
 /**
  * A very simple wrapper for an SVG icon required from the static repository.
- * NOTE! In v2 the only way to use icons in back-bone ie. without react is to require using a different loader
- * (for webpack) - you *must also use the leaging "!" in your loader so the default SVG->React loader is not used.
+ * NOTE! In v2 the only way to use icons in backbone ie. without react is to require using a different loader
```
walfly: this isn't actually true, right? You can require the dist/templateString/*.js files without a loader as they are just plain js.

hellatan: Idk. I actually didn't read what it said as this was originally there. @nameofname would know more about this. I was only fixing typos 😀

nameofname: @walfly is correct.

hellatan: i'm leaving this in this PR. if we need to remove it, then it can be removed in a subsequent PR since this was just fixing typos.

hellatan:
@nameofname @walfly other than the typo comments, anything else? I also tested this so that the index.html file will still work as a standalone page without needing webpack dev server running.

js/SvgStringToReact.jsx
```diff
@@ -1,29 +1,30 @@
...
-    }
-});
+const SvgStringToReact = props => {
+    const styles = {
+        fill: "currentcolor",
+        verticalAlign: "middle"
+    };
+    return (
+        <svg
+            viewBox="0 0 24 24"
```
nameofname: looks like this will give all the icons the same viewBox. This shouldn't be true since they are all different.

js/SvgWrapper.js
nameofname: @walfly is correct - can you just delete this comment.

js/SvgStringToReact.jsx
hellatan: this was originally there. i believe @walfly wrote this one. I only converted this to use `extends React.Component` instead of `React.createClass`. This should be updated in a different PR.

hellatan: err, i made this into a stateless component and got rid of `React.createClass`. Everything else about it is from the original code.

js/SvgWrapper.js
hellatan: ok - that i can do :)

js/SvgStringToReact.jsx
walfly: This isn't being used.  I added it because i was confused based on the documentation.  It should be removed.  I'll open a new pr that gets rid of it.

sample/js/BackboneIcons.js
```diff
@@ -4,31 +4,33 @@
...
  * Renders each of the icons using the svg wrapper backbone view.
  */
 const BackboneIcons = Backbone.View.extend({
 
-    allIcons() {
-        return iconList.map((icon, idx) => {
+    _allIcons() {
+        return iconList.map(icon => {
             // NOTE* Here we have to use a leading "!" to tell webpack not to use the svg->react loader
             const Icon = require('!html!../../src/' + icon);
```
nameofname: if you want you can get rid of this !html loader thing here and just use the strinigfied version in the dist folder instead.

nameofname:
@hellatan this is cool. I'm looking forward to seeing what it looks like now. Once this merged in please also merge into the gh-pages branch and push to upstream which will update this page : beautifulcode.1stdibs.com/dibs-vg/

js/SvgStringToReact.jsx
walfly: https://github.com/1stdibs/dibs-vg/pull/33

nameofname:
It's merged, please publish @hellatan

hellatan:
published:

> \+ dibs-vg@1.9.0

hellatan:
@nameofname @walfly - gh-pages updated: http://beautifulcode.1stdibs.com/dibs-vg/

PR #34, opened 2016-07-19
Updates

hellatan:
gonna merge this tonight

@walfly @nameofname - I added a "gotchas" section to the readme. Apparently, the compile step does not like the `xml:space="preserve"` property but it does not error out.


hellatan:
> \+ dibs-vg@1.9.1

PR #36, opened 2016-07-26
Replace HTML/SVG attributes to React attributes

gedas-monginas:
At the moment any attributes inside `<svg` don't get modified.
Currently we have some SVG's with `class` and `fill-rule` attributes, so added a babel plugin which converts just those.
Plan is to make a PR on https://github.com/insin/babel-plugin-react-html-attrs and if its accepted use that plugin instead.
screen shot 2016-07-26 at 15 13 08


kostasmanionis:
I'll open up a PR there to add more attributes.

src/person-inside-badge.svg
```diff
@@ -1,33 +1,7 @@
...
-                    c4.4-31-11.7-39-13.6-39.8c-5.3-2.7-10.9-4-16.7-4c-9.9,0-16.9,4-17.2,4.2C100.8,67.5,96.3,75.8,95,87c-0.9,7.9,0.1,14.6,0.2,15.3
-                    c1.9,14.1,10.2,23.8,10.5,24.2l0.1,0.1c1.2,1.3,1.2,1.3,1.6,2.2c0.1,0.3,0.1,1.2,0.1,1.9c-1,4.6-4.9,6.2-5.1,6.2l-8.8,3.8
-                    l-20.8,8.4c-6.8,2.4-9.6,8.7-10,9.6c-2.5,5.5-3.5,20.2-3.6,21.9l-0.2,3l5.7-0.1c0.1,0.1,0.3-2.4,0.3-4l0.1-2.6
-                    c0.5-5.7,1.4-12.7,2.9-15.7l0.1-0.3c0,0,1.9-4.8,6.6-6.5l21-8.5l8.8-3.8c0.9-0.3,6.9-3,8.5-10.4l0-0.2c0.1-0.8,0.4-3.4-0.6-5.3
-                    c-0.8-1.6-1-1.8-2.5-3.5c-0.5-0.6-7.6-9.3-9.2-21.3l0-0.1c0-0.2-3.9-24.9,10.4-33.5c0.1,0,5.9-3.4,14.3-3.4
-                    c4.9,0,9.7,1.2,14.2,3.4l0.2,0.1c0.6,0.2,14.2,6.2,10.2,33.9l0,0.1c0,0.1-1.4,11.6-9.2,20.9c-0.1,0.2-0.3,0.3-0.5,0.5
-                    c-0.7,0.8-1.7,1.8-2.3,3.1l-0.1,0.2c-0.1,0.3-1,2.6,0,5.5c0.3,1.2,2,6.8,9.2,10l11.4,5.1l17,6.7c0.3,0.1,7.3,3,9,12.7l0.8,11.1
-                    c0,0,0,0,0,0c0,0,0.2,5,0.4,5H191.3z"/>
-        </g>
+    <g fill-rule="evenodd">
```
hellatan: What's "evenodd"?

Also, what all changed in this diff? Hard to really tell since it was multiline but now the paths are all single line.

gedas-monginas: this `fill-rule="evenodd"` was already here and @moriko probably can better explain what is used for and what a the changes, as much as I know there were bunch of unused paths left and icon should look identically.

moriko: > What's "evenodd"?
This value determines the "insideness" of a point in the shape by drawing a ray from that point to infinity in any direction and counting the number of path segments from the given shape that the ray crosses. If this number is odd, the point is inside; if even, the point is outside.

> Also, what all changed in this diff?

Made the svg cleaner as it had unnecessary paths.

kostasmanionis: Fancy :D Any idea how many more unclean svgs do we have?

gedas-monginas: there is no other way to tell just to go through all of them :)

moriko: From time to time I do come across with some dirty ones. Whenever will have time for the cleanup will do the sweep :D

hellatan: how can you tell if it's a "dirty" svg?

moriko: > how can you tell if it's a "dirty" svg?

First by how the svg looks and how it corresponds to the path/polygons/rectangles that could describe it. Next I just check it in Sketch. This svg had two same circles, one dot that wasn't even visible and some grouping that was not needed in any way.

htmlToReactAttributes.js
```diff
@@ -0,0 +1,20 @@
...
+const ATTRIBUTES = {
+    'class': 'className',
+    'fill-rule': 'fillRule'
+};
+
+const transformHtmlToJsxAttributes = () => {
+    return {
+        visitor: {
+            JSXAttribute: function (node) {
+                if (node.node.name.name in ATTRIBUTES) {
```
nameofname: Any chance node.node.name.name would not be defined in this context? If so please add some type checking.

src/person-inside-badge.svg
nameofname: Dang! Nice job. When I was importing these all and editing by hand I noticed that there were some dirty ones as well. This happens depending on the way the designer constructs their file in Illustrator, ie. if they have an extra path or layer, Illustrator will create additional elements. Of course no pressure to clean those up if they look ok.

htmlToReactAttributes.js
kostasmanionis: The first `node` probably should be named `path`. And the `name.name` is weird, but it seems that's how jsx attributes are constructed in AST 😕 

screen shot 2016-07-26 at 18 37 33

So I think it's pretty safe

gedas-monginas:
any objections on merging this? and should this be a major or minor?

hellatan:
i would say minor. No objections here - @nameofname?

nameofname:
@gedas-monginas go ahead.

PR #45, opened 2016-09-12
[89] FINDING-8963 FINDING-8964 : added add-tags.svg

CarolineHo:
Added add-tags.svg

dibs-cms-trade PR: https://github.com/1stdibs/dibs-cms-trade/pull/229


hellatan:
maybe just call this `add-tags` instead? does that make more sense?

CarolineHo:
@hellatan changed name to `add-tags`

hellatan:
published:
```
+ dibs-vg@1.16.0
```

PR #46, opened 2016-09-27
add enabler

walfly:


nameofname:
A couple things here : 

In general we don't use fill for our SVGs - but since this one is a color picture, it needs to use fill. Since this SVG is so different can you put it in a special dir for pictures like this? 

Also - we always use the class name dibs-vg, and a special unique class name, please see example here : https://github.com/1stdibs/dibs-vg/blame/master/src/baby-bottle.svg#L1, and copy that.

nameofname:
@walfly what's the deal with this PR -- for the moment I am thinking this SVG really belongs in your enable repository - not like anybody else is going to need to use it, plus it's thousands of lines long and it's going to bloat an otherwise small repository. Closing, re-open if you feel strongly.

PR #51, opened 2016-11-10
FINDING-9453 added payment methods icons

untitledlt:


src/payment-american-express.svg
```diff
@@ -0,0 +1,12 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-american-express" viewBox="0 0 500 312" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <rect class="dibsvg-color-holder" width="500" height="312" fill="#F8F8F8"/>
```
hellatan: does the `fill` attribute override any fill from css?

src/payment-american-express.svg
```diff
@@ -0,0 +1,12 @@
...
+    <g class="dibsvg-color-letters" fill="#2E77BC" transform="translate(7.028 73.293)">
```
hellatan: same here re:fill attribute

src/payment-check.svg
```diff
@@ -0,0 +1,4 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-check" viewBox="0 0 500 312" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <rect class="dibsvg-color-holder" width="500" height="312" fill="#F8F8F8"/>
```
hellatan: same here re:fill attribute

src/payment-discover.svg
```diff
@@ -0,0 +1,13 @@
...
+        <path d="M14.9048096,1.75350701 L0.292251169,1.75350701 L0.292251169,57.0724783 L14.9048096,57.0724783 C23.254843,57.0724783 28.4736139,54.9849699 33.6923848,50.8099532 C39.9549098,45.5911824 43.0861723,37.241149 43.0861723,29.9348697 C44.1299265,13.2348029 31.6048764,1.75350701 14.9048096,1.75350701 L14.9048096,1.75350701 L14.9048096,1.75350701 Z M27.4298597,43.503674 C24.2985972,46.6349365 20.1235805,47.6786907 12.8173013,47.6786907 L10.7297929,47.6786907 L10.7297929,11.1472946 L13.8610554,11.1472946 C20.1235805,11.1472946 24.2985972,12.1910488 28.4736139,15.3223113 C31.6048764,18.4535738 33.6923848,23.6723447 33.6923848,28.8911156 C32.6486306,35.1536406 30.5611222,40.3724115 27.4298597,43.503674 L27.4298597,43.503674 L27.4298597,43.503674 Z"/>
+    </g>
+    <path class="dibsvg-color-circle" fill="#F6821F" d="M0,31.3126253 C0,48.0126921 12.5250501,60.5377422 28.1813627,60.5377422 C43.8376754,60.5377422 56.3627255,48.0126921 56.3627255,31.3126253 C56.3627255,15.6563126 43.8376754,3.13126253 28.1813627,3.13126253 C12.5250501,3.13126253 0,15.6563126 0,31.3126253 L0,31.3126253 Z" transform="translate(228.624 112.809)"/>
```
hellatan: and another - same here re:fill attribute

hellatan:
It looks like all your svgs have the `fill` attribute. Make sure that they can be overwritten by css. If not, then remove that attribute.

src/payment-american-express.svg
untitledlt: those are color icons so we should not modify them with css.
cc @moriko

src/payment-check.svg
hellatan: you should remove this one though b/c it's not a branded logo

src/payment-wire-transfer.svg
```diff
@@ -0,0 +1,16 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-wire-transfer" viewBox="0 0 500 312" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <rect class="dibsvg-color-holder" width="500" height="312" fill="#F8F8F8"/>
```
hellatan: you should also remove the fills here since this is a non-branded icon as well

hellatan:
any updates on this?

untitledlt:
working on it

untitledlt:
@hellatan looks like it's possible to overwrite `fill` with `css`.
All elements with `fill` has class name so it is possible to style any of them.

hellatan:
i think it's still cleaner to remove that attribute when it's not necessary.

untitledlt:
@hellatan those payment icons has grey background and if we do not set `fill` attribute, they will be black by default. So one way would be to set that background color in `head.css` or set it with `fill` attribute. Or do you have a better idea?

hellatan:
i meant the non-branded ones that don't rely on the multiple colors

hellatan:
oh you mean this:
```
<rect class="dibsvg-color-holder" width="500" height="312" fill="#F8F8F8"/>
``
is the background?

**Comment by untitledlt** [comment]
yep! it's used in brand icons and in non-branded

**Comment by hellatan** [comment]
published:
```
+ dibs-vg@1.19.0
```

PR #53, opened 2016-11-17
[99] FINDING-9569: added tag.svg

CarolineHo:


hellatan:
this looks kind of weird how the tag doesn't butt up against the right edge:

screen shot 2016-11-23 at 1 12 55 pm

hellatan:
can you ask whoever created this svg to crop it so that the tag butts up right against the right edge? Should this maybe be a vertical tag and then css is applied to it to rotate it instead?

lilianammmatos:
@CarolineHo added `don't merge` label while you get the new SVG.

CarolineHo:
@hellatan @lilianammmatos just uploaded the new svg

hellatan:
i can't remember - but did we want everything to butt up against the edges:

screen shot 2016-11-29 at 2 41 21 pm

hellatan:
⬛️ is the answer

hellatan:
@CarolineHo please version and publish. This should be a minor.

CarolineHo:
published dibs-vg@1.22.0
(for this PR: https://github.com/1stdibs/dibs-vg/pull/56)

PR #57, opened 2016-11-30
default fill for two color svg

hellatan:
if we have a multi-color svg, there should be default fills on the elements we are targeting in addition to classes so we can also affect the fill via css. 

without fill:
screen shot 2016-11-30 at 12 26 37 pm

with fill:
screen shot 2016-11-30 at 12 27 47 pm

also, this should be called `logo-mobile` to keep consistency with the current logo naming convention. Since that would be a breaking change, that will be updated in the next major rev.

cc @cormickjbrowne


hellatan:
@nameofname updated stuff we discussed offline

hellatan:
@nameofname oh wait, need to update a couple more svgs with classes

hellatan:
@nameofname ok all set now

package.json
```diff
@@ -34,10 +34,6 @@
     "url": "https://github.com/1stdibs/dibs-vg/issues"
   },
   "homepage": "https://github.com/1stdibs/dibs-vg#readme",
-  "dependencies": {
-    "backbone": "^1.3.3",
-    "jquery": "^2.2.3"
-  },
   "devDependencies": {
```
nameofname: @hellatan - it just occurred to me that we should probably have react as a peer dep. You agree?

nameofname:
@hellatan weren't you going to change a file name too?

hellatan:
peerDep - yes. file name change - no, that's going to be in the major rev

hellatan:
@nameofname - peerDep is actually already there: 
```json
  "peerDependencies": {
    "react": "^15.0.2",
    "react-dom": "^15.0.2"
  }
```

hellatan:
so this is good to merge whenever

hellatan:
> + dibs-vg@1.22.1

Issue #39, opened 2016-07-31
Update for svg-react-loader@v0.4.0


jhamlet:
Hi there,

Just came across your npm module, and your repository.

I want to let you know that I am getting ready to launch a v0.4.0 version of `svg-react-loader` that is a complete refactor. It no longer needs `babel` and opens up an API that allows you to fully customize the final SVG component output to `React`.

You can review the update here: https://github.com/jhamlet/svg-react-loader/blob/version-0.4.0/README-v0.4.0.md and you can install it with `npm install svg-react-loader@next`

I would appreciate any feedback you have since you did catch one particular `webpack` configuration step I neglected in my own documentation.

Thanks,

;-j

hellatan:
hey @jhamlet! I just came across this issue. I'll take a look and let you know if anything comes up. It looks like 0.4.0 hasn't been published yet?

jhamlet:
@hellatan officially no. You can get it by doing a `npm install svg-react-loader@next` or `npm install svg-react-loader@0.4.0-beta.2` if you want to be precise.

hellatan:
@jhamlet is there anything in particular i should look out for? I'm not sure which webpack issue you're talking about but I'll talk to my co-worker @nameofname as well, as he's the one that set things up originally.

I just installed `svg-react-loader@0.4.0-beta.2` and this is the result of a file that did not use babel to transform its contents:

```js
const React = require('react');
const Component = React.createClass({
    render: function() {
        return (
            <svg className={ [this.props.className, 'dibsvg-svg dibsvg-account-filled'].join(' ')} viewBox="0 0 500 500">
[...]

PR #59, opened 2016-12-08
Remove jsx

hellatan:
this will be a major bump


hellatan:
also, look at https://github.com/1stdibs/dibs-vg/issues/39#issuecomment-265569259. Updating the svg-react-loader changed the `svg` output in the react component.

hellatan:
didn't do this soon enough. And I really don't want to major bump this in all the deps we have now so going to close.

PR #60, opened 2017-01-19
added iim svg icons

sashak007:


hellatan:
we should really reuse the arrows we already have. Also, you should have only one arrow and then you can rotate it with styles instead. The original go with the multiple arrows in different directions was the incorrect approach.

hellatan:
i feel like a lot of these icons we already have, they are just styled differently.

- instead of pencil, we have a pen - https://github.com/1stdibs/dibs-vg/blob/master/src/pen.svg
- instead of time, there's a clock already - https://github.com/1stdibs/dibs-vg/blob/master/src/clock.svg
- there's a grid icon - https://github.com/1stdibs/dibs-vg/blob/master/src/grid.svg
- list icon - https://github.com/1stdibs/dibs-vg/blob/master/src/list.svg

There's an overlay spinner package that can be used in lieu of an icon, thought that one is all css.

Lol, the "chevron-down" icon is the exact same one as the "arrow-down" icon in this PR. Really, all the arrow and chevron icons shouldn't be apart of this PR.

Product should communicate on their icon sets b/c there is a lot of overlap. If these icons need to be  very specific to IIM, then i would propose making a different icon set than putting it in this one.

In the mean time, i'm going to close this PR until those product decisions have been clarified.

nameofname:
@hellatan @sashak007 - can we re-open this with only the SVGs that are not arrows? I think Dale is right, let's just use the arrows that we have and rotate them, but I would love to have the other IIM icons that we have here in SVG format - like the download link, zoom, thumbnail, etc.

list.json
```diff
@@ -1 +1 @@
...
...vg","info-circle.svg","instagram.svg","iphone.svg","lightbulb.svg","linkedin.svg","list.svg","loader.svg","location-pin.svg","lock-with-checkmark.svg","lock.svg","logo-1st.svg","logo-introspective.svg","logo-no-reg.svg","logo.svg","luggage.svg","magnifying-glass.svg","mask.svg","message.svg","minus-sign-inside-a-black-circle.svg","minus.svg","mobile-logo.svg","net-price.svg","orange.svg","payment-american-express.svg","payment-check.svg","payment-discover.svg","payment-mastercard.svg","payment-paypal.svg","payment-visa.svg","payment-wire-transfer.svg","pen.svg","pencil.svg","person-inside-badge.svg","person-profile.svg","pineapple.svg","pinterest.svg","plus.svg","print.svg","quote.svg","ramen.svg","reject.svg","repost.svg","return-arrow.svg","service-person.svg","share.svg","shipping-package.svg","shipping-world.svg","shipping.svg","shopping-cart.svg","spinner-of-dots.svg","star-filled.svg","star-outlined.svg","storefront.svg","tag.svg","tax-free.svg","thumbnail-grid.svg","thumbnail-list.svg","time.svg","toolbox.svg","trade-arrow.svg","trade-logo.svg","trash-filled.svg","triangle-down.svg","truck-moving.svg","twitter.svg","upload-failed.svg","yoga.svg","zoom-in.svg","zoom-out.svg"]
```
hellatan: you might have an outdated repo. This should have very specific diffs like this - https://github.com/1stdibs/dibs-vg/pull/72/files#diff-2a091062a700008c2c174a1edd354d92

hellatan: oh this was reopened from an older PR so that makes sense that you would have this file 😄

src/thumbnail-list.svg
```diff
@@ -0,0 +1,22 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-thumbnail-list" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 511.626 511.627">
```
nameofname: kill this one.

src/chevron-down.svg
```diff
@@ -0,0 +1,7 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-chevron-down" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 444.819 444.819">
```
nameofname: we're removing all the chevrons

src/pencil.svg
```diff
@@ -0,0 +1,8 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-pencil" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 432.544 432.544">
```
nameofname: use the pen icon instead

src/thumbnail-grid.svg
```diff
@@ -0,0 +1,31 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-thumbnail-grid" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 511.626 511.627">
```
nameofname: use the 2 X 2 grid instead.

src/exclamation-sign.svg
```diff
@@ -0,0 +1,12 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-exclamation-mark" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 438.533 438.533">
```
nameofname: as discussed offline, we are going to try to get the current exclamation mark we use inside of a circle, instead of using this one which has different styling.

src/time.svg
```diff
@@ -0,0 +1,15 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-time" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 438.533 438.533">
```
nameofname: Also just saw @hellatan's old comment - we can ditch the time icon in favor of the clock icon that's already in there.

src/download-symbol.svg
```diff
@@ -0,0 +1,16 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-download" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 475.078 475.077">
```
nameofname: This one can go too - we have download.svg, don't need download-symbol.svg.

src/close-circle.svg
```diff
@@ -0,0 +1,14 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-close-circle" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 438.533 438.533">
```
nameofname: Can we also get the current close icon inside of a circle? This one is nice but just realized it's not going to match at all with the other close icon.

src/exclamation-sign.svg
sashak007: as discussed offline, waiting for assets from Jen.

src/close-circle.svg
sashak007: as discussed offline, waiting for assets from Jen.

src/minus-sign-inside-a-black-circle.svg
```diff
@@ -0,0 +1,10 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-minus" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 438.533 438.533">
```
moriko: We already have `minus` icon so `dibsvg-minus` should be renamed. And suggest naming this icon `circle-minus` or `minus-circled`. Or for future reference even better would be that all those circled in dark background would be named accordingly `circle-bold-*`. But after reading other comments I believe that these darkish icons will be removed anyway.

Also the viewport for the most svg icons is `0 0 250 250`. If this is different it's a hint that the designer didn't  follow the guidelines.

nameofname: @moriko you are right re: view port, however we don't currently enforce view port dimensions - just as long as it's square - since the developer will have to resize it anyways. 

Thanks for catching the thing about the naming - this should definitely be "minus-circle" or  "minus-circle-filled" to follow conventions.

moriko: Sometimes the resizing of viewport can break the svg icon so should be a bit careful with that.

nameofname: Can you elaborate on that or point me to a resource? If that's true we should re write parts of the documentation (I just updated) - here : 

https://github.com/1stdibs/dibs-vg/blob/master/CONTRIBUTING.md

I'm fine with amending our requirements if it serves a purpose.

moriko: Lets take svg icon `close-circle` and change the viewport to smaller square size:

previous `438.533`  | changed to `250`
--------- | ---------
screenshot 2017-04-05 17 37 25 | screenshot 2017-04-05 17 38 03

nameofname: @moriko thanks for elaborating. IMO it's not an issue because the same would be true for any SVG. Basically the view port should match the bounding box of the SVG path and the developer should never modify the view port attribute. 

Hope this alleviates your concern.

hellatan:
we really need to sync up with product on these icons. The strokes/weight on some of these icons are conflicting with the information i'm getting.

PR #67, opened 2017-02-23
Contemporary svgs

cormickjbrowne:


.eslintignore
```diff
@@ -0,0 +1 @@
+sample/compiled
```
hellatan: you don't want this linted?!?!? 😂

src/trashcan.svg
```diff
@@ -0,0 +1,5 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-trashcan" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
```
hellatan: there's a "trash-filled" icon. If we are going to go with a non-filled one, we should make the trash cans line up in design. I prefer this style tbh.

.eslintignore
cormickjbrowne: The linter throws ~30k errors and ~50k warnings. I thought it would be ok to ignore it. The linter still throws a warning and requires a -n flag now that I'm committing a file that is lint-ignored.

nameofname: Yes - ignore it. No reason to lint the compiled JS.

hellatan: obviously joking, dummies 😛

src/trashcan.svg
hellatan: looks like this is only for enabler stuff - https://github.com/search?q=trash-filled+org%3A1stdibs&ref=searchresults&type=Code&utf8=%E2%9C%93 - We should update that one though. If it looks messed up on an enabler page, i'm not as concerned about just changing it now

cormickjbrowne: Should we add the new icon now and then open a ticket to switch the icons and remove the old svg?

hellatan: modify the current one and rename this one "trash-empty"

cormickjbrowne: Seems like the difference between a trash-empty and trash-filled icon would be accomplished by style. Do we need two components for that?

hellatan: i guess this name is fine then, but use a hyphen - "trash-can". the enabler trash can is always filled right now and only changes color. We should change those at some point to use the trash-can instead.

cormickjbrowne: Created a ticket to clean this up: https://1stdibs.atlassian.net/browse/DX-1797

jrdrg:
+ dibs-vg@1.28.0

PR #68, opened 2017-02-23
DX-1778 icons for dealer analytics page

pearfish:
needed for https://github.com/1stdibs/app-dealer-tools/pull/295


hellatan:
@pearfish @cormickjbrowne are all of your icons only for DX UI stuff? I'm wondering if we should maybe make a separate icon set or do we want to continue growing just this one? @nameofname thoughts?

pearfish:
for now these should all be only used on DX apps yeah, I'm not sure if design has any long term plans for these icons though

hellatan:
just a thought. i guess if we want consistent UI elements between apps, it shouldn't matter the app context, just the component context, i suppose

jrdrg:
ping @hellatan @nameofname does this look ok to you guys? Wanna merge this so it doesn't block other stuff for 109

jrdrg:
+ dibs-vg@1.28.0

PR #70, opened 2017-03-01
DX-1698 Add play button svg

corbinschwartz:


src/play-button.svg
```diff
@@ -0,0 +1,5 @@
+<svg viewBox="0 0 100 100" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/2000/xlink" class="dibsvg-svg dibsvg-play-button">
+    <g id="play-button">
```
hellatan: do you need this `id`?

corbinschwartz: Nope. The svg started as something exported from Sketch, I had to modify it a decent amount to get it to work how we like. Thanks for catching the id

jrdrg:
+ dibs-vg@1.30.0

PR #71, opened 2017-03-22
GOODS-5932 Add eye svg

aahill50:
:eye:


src/eye.svg
```diff
@@ -0,0 +1,26 @@
...
+	<g>
+		<path class="st0" fill="none" stroke="#000" stroke-width="2" stroke-miterlimit="10" d="M93.9,264.7c0,0,151.6-236.5,330.9,2.6"/>
+		<g>
+			<defs>
+				<ellipse id="SVGID_1_" cx="259.4" cy="253.9" rx="48.2" ry="49.4"/>
+			</defs>
+			<clipPath id="SVGID_2_">
+				<use xlink:href="#SVGID_1_"  style="overflow:visible;"/>
+			</clipPath>
+			<rect x="146.5" y="139.9" class="st1" clip-path="url(#SVGID_2_)" fill="#000000" width="225.7" height="228"/>
```
hellatan: i don't think `clip-path` can be used - http://caniuse.com/#search=clip-path - IE doesn't have any support for that property. Also, remove any `fill` property. If you need to give it a different color, you should create a custom class for it. I think the designer may need to figure out a different way to do this "eye" - look at the "bullet-point" svg

hellatan:
a lot of this svg looks strange. there are a bunch of `stroke`, `stroke-mitersomething`, etc attributes that we don't typically have for any other svg. @nameofname can you sort out what we can and can't do with this? It looks like it was taken directly from sketch without any modifications.

aahill50:
@hellatan are there guidelines that the designers should be following when they give us these svgs? I also noticed that it was constructed differently than the other ones I had seen.

hellatan:
> are there guidelines that the designers should be following when they give us these svgs? 

from a design standpoint yes, but it's hard to dictate that from an output point of view since all designers see is their svg in sketch or whichever app they are using. I guess one thing we can tell them is to minimize the number of folders/groups they use so it removes a bunch of `<g>` elements

src/eye.svg
nameofname: I agree - the SVG should not be using stroke. To do that you have to send this back to a designer. Any IDs or class names automatically added by adobe must be removed as well. 

Honestly the best way to clean this up is to kick it back to a designer, have them outline, flatten and merge all paths together, then send it back to you. Jen C. is really good at that.

moriko:
@aahill50 @hellatan I've created similar SVG that should be ok for use:
screenshot 2017-03-27 16 45 39

Path to prepared to upload SVG:  https://www.dropbox.com/s/eti5hmdlqc05tlx/eye.svg?dl=0

hellatan:
@moriko i'm not seeing it here: https://qa.1stdibs.com/style-guide/iconography/ and i don't remember seeing it here (currently down): http://beautifulcode.1stdibs.com/dibs-vg. The closest one i saw was the "top-view" icon, but was slightly different.

moriko:
@hellatan I JUST created this icon in Sketch and leaving it for you guys to upload :)

hellatan:
@moriko ah, gotcha

hellatan:
@nameofname what were our rules for the viewbox? did we want to keep as many icons as possible a square so there would be known whitespace?

aahill50:
I think we usually request the viewBox to be square, @nameofname can confirm

nameofname:
@aahill50 yup. square. I think there are one or two exceptions, but for icons especially they should always be square to avoid layout issues.

nameofname:
@hellatan didn't we have a document that described the steps designers were supposed to take when creating an SVG from Illustrator? Those rules should still be in effect to avoid this kind of thin in the future. I know Jen follows the correct steps - let's check in with design team on this later today.

hellatan:
in lieu of @moriko updating the version she made, @nameofname or @aahill50 can one of you get jen to update it to be square?

hellatan:
^^ i'd just give jen @moriko's version and update from there

moriko:
@dale if viewBox will be changed to `viewBox="0 -65 350 350"` it will be square and centered.

aahill50:
@moriko Thanks, that did the trick!

hellatan:
ugh...forgot to hit submit earlier. I'd prefer to NOT have these arbitrary coordinates in the viewBox attribute. Then we'll get in the habit of nudging some icons this way and some icons that way when we should really fix it at the source so all the icons are in a consistent coordinate system.

aahill50:
Ok finally got the designers to fix thing up. I believe this should be good to go

aahill50:
Published `dibs-vg@1.32.0`

PR #72, opened 2017-03-23
DX-1863 Adds x svg (#71)

infinityfarm:
@jrdrg @nameofname @hellatan


src/x.svg
```diff
@@ -0,0 +1,7 @@
+<svg class="dibsvg dibsvg-x" viewBox="0 0 22 22" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
```
hellatan: don't use the `stroke*` attributes - it's not x-browser compatible. I would also get rid of this `<g>` tag (usually means it's nested in sketch). @nameofname mentioned that jen cassel is good at cleaning up icons for us

src/x.svg
```diff
@@ -0,0 +1,7 @@
...
+        <g transform="translate(-970.000000, -1212.000000)" fill-rule="nonzero" fill="#000000">
```
hellatan: remove the `fill*` attributes. If you really need to have a separate fill color, add a class and target the class. There are exceptions when it's a specific brand (ie. paypal, credit card icons, etc).

I also feel like the "x" can probably be drawn so that it doesn't need to be translated (sounds like it was rotated). I'm not 💯  sure though.

src/x.svg
```diff
@@ -0,0 +1,7 @@
...
+            <polygon id="x-svg" points="989.20662 1212.58501 991.384268 1214.74245 983.170654 1223.05623 991.414462 1231.2063 989.250218 1233.39232 981.008033 1225.23682 972.930866 1233.41534 970.745029 1231.2496 978.828544 1223.07998 970.588585 1214.9343 972.739508 1212.7414 980.985104 1220.89382"></polygon>
```
hellatan: do you need this `id`?

nameofname: @hellatan the ID _has to_ be removed. Adobe puts IDs in the markup all the time, but they can be the exact same ID from one file to another. Not safe to keep IDs that adobe added to your markup.

jrdrg:
+ dibs-vg@1.31.0

PR #73, opened 2017-03-30
Update readme

nameofname:
Added a contributing MD file. 
Added the contributing MD file to the sample page. 

@hellatan I thought this would be a good place to share instructions on contributing, but I could do without it too. Let me know what you think. It's a public repo so designers can view the normal CONTRIBUTING.md file.


sample/js/Wrapper.jsx
```diff
@@ -13,6 +13,9 @@ const styles = require('../scss/sample.scss');
 const Wrapper = () => {
     return (
         <div>
+            <div className={styles.nav}>
+                <a href="https://github.com/1stdibs/dibs-vg/blob/master/CONTRIBUTING.md">Guide To Contributing</a>
```
hellatan: @nameofname can you post a screenshot of what this will look like in the gh-pages build?

CONTRIBUTING.md
```diff
@@ -0,0 +1,58 @@
+# Contributing
+
+## Instructions on creating SVG files via Illustrator
```
hellatan: @nameofname are there sketch instructions? I think i've seen some svgs exported by sketch.

nameofname: No - I have seen one SVG coming from sketch, we don't have instructions for it yet, we can add later if someone knows ... anything about that.

sample/js/Wrapper.jsx
nameofname: screen shot 2017-03-30 at 4 07 33 pm

PR #78, opened 2017-05-04
add subsection to CONTRIBUTING.md regarding interior elements of svgs

orteth01:
I'm no SVG expert so I just added a subsection relating to the `fill` attribute, which is what tripped me up in my most recent PR. 

As we come across more attributes that can be removed from interior elements we can easily add them. @nameofname if you know any off the top of your head, let me know and I'll add them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,19 @@ This folder includes a master illustrator file, and a folder for each icon indiv
 			- The `xmlns` and `xmlns:xlink` attributes are for XML name spacing and aren't strictly necessary, but they are useful for displaying SVGs outside of the context of HTML documents, so we like to include them. [Reference](http://stackoverflow.com/a/18468348/4002508), [Explanation](https://developer.mozilla.org/en/docs/Web/SVG/Namespaces_Crash_Course)
 			- `version` is not required, and is ignored by browsers but we currently don't force developers to delete it.
 	- interior elements (e.g. `<path />`, `<polygon />`, etc.)
-		- remove the `fill` attribute if the element's fill color is intended to be black (e.g. `#000000`, `#010101`). The default value for `fill` is black so this attribute is unnecessary in these cases.
+		- remove the `fill` attribute if the element's fill color is intended to be black (e.g. `#000000`, `#010101`). The default value for `fill` is black (`#000000`) so this attribute is unnecessary in these cases.
+        - remove `<g />` elements that have no attributes and aren't actually being used for grouping separate parts of the SVG.
+            - SVGs generated from illustrator sometimes look like this:
+                ```
+                <svg ... >
+                    <g>
+                        <path ... />
+                        <polygon ... />
+                        etc....
+                    </g>
+                </svg>
+                ```
+                the `<g>` tags are unnecessary and can be removed.
 
 5. Remove any classes or IDs illustrator added to the SVG elements. For example, sometimes you will see markup that looks like this : 
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,13 +46,13 @@ This folder includes a master illustrator file, and a folder for each icon indiv
 
 3. Add class names to your icon using the format "dibsvg-svg dibsvg-{FILE_NAME}", where FILE_NAME is the name of the file you created.
 
-4. Clean up the SVG file by removing other unnecessary, auto-generated properties:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -46,9 +46,13 @@ This folder includes a master illustrator file, and a folder for each icon indiv
 
 3. Add class names to your icon using the format "dibsvg-svg dibsvg-{FILE_NAME}", where FILE_NAME is the name of the file you created.
 
-4. Clean up the SVG file by removing other auto generated properties on the SVG element. The only allowed properties at this time are : class, viewBox, xmlns, xmlns:xlink, and version. 
-	- The xmlns and xmlns:xlink attributes are for XML name spacning and aren't strictly necessary, but they are useful for displaying SVGs outside of the context of HTML documents, so we like to include them. [Reference](http://stackoverflow.com/a/18468348/4002508), [Explanation](https://developer.mozilla.org/en/docs/Web/SVG/Namespaces_Crash_Course)
-	- Version is not required, and is ignored by browsers but we currently don't force developers to delete it. 
+4. Clean up the SVG file by removing other unnecessary, auto-generated properties:
+	- `<svg />` element
+		- The only allowed properties on the `<svg />` element at this time are : class, viewBox, xmlns, xmlns:xlink, and version.
```
hellatan: can you wrap the attribute names in backticks to give code highlighting?

orteth01:
@nameofname  @hellatan 

can empty `<g>` tags like [this](https://github.com/1stdibs/dibs-vg/blob/master/src/bullet-point.svg?short_path=601bacb#L2-L6) be removed?

i see a few SVGs like this and the `<g>` tag seems unnecessary...

hellatan:
> can <g> tags with no attributes, like this, be removed?

I think they can if they are not grouping multiple elements.

nameofname:
@orteth01 re: <g> tags yes they can be removed - however it's not necessary by any means. Worth calling out if you want to add it.

orteth01:
@hellatan @nameofname okie dokie i've added a note about `<g />` tags

CONTRIBUTING.md
```diff
@@ -46,9 +46,25 @@ This folder includes a master illustrator file, and a folder for each icon indiv
...
+		- remove the `fill` attribute if the element's fill color is intended to be black (e.g. `#000000`, `#010101`). The default value for `fill` is black (`#000000`) so this attribute is unnecessary in these cases.
+        - remove `<g />` elements that have no attributes and aren't actually being used for grouping separate parts of the SVG.
+            - SVGs generated from illustrator sometimes look like this:
+                ```
+                <svg ... >
+                    <g>
+                        <path ... />
+                        <polygon ... />
+                        etc....
+                    </g>
```
hellatan: so i think with this structure, the `<g>` is needed. If it's like this:
```
<svg ...>
    <g>
        <path ... />
    </g>
</svg>
```
then you'd want to take it out since it's a group of 1, not a group of many.

nameofname: @hellatan I thought the only purpose of a <g> tag in markup is to apply transformations to many grouped elements. Do you know if grouping elements impacts the stacking order? eg. is it possible to break the appearance of an SVG by removing <g> tags?

orteth01: @hellatan I still think it can be removed regardless of it has one or many interior elements.

here's  one of our icons that has many elements contained within a `<g>` and then with the `<g>`s removed....
https://jsfiddle.net/4xqk5toe/

hellatan: i guess if there are no transformations, it doesn't matter. We should try to not have any transformations as well if we can avoid it. Stacking order, I'm not really sure.

orteth01: it would blow my mind if removing a `<g>` tag that has no attributes would somehow affect stacking order... looking into it though...

orteth01: from the [mozilla docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/g)

> The <g> SVG element is a container used to group other SVG elements. Transformations applied to the <g> element are performed on all of its child elements, and any of its attributes are inherited by its child elements. It can also group multiple elements to be referenced later with the <use> element.

i think it is safe to assume stacking order is unchanged by le `<g>`... WDYT?

hellatan: > WDYT
LGTM

PR #84, opened 2017-05-26
MG-1018 add data tags for image icons

szhu1026:


scripts/htmlToReact.js
```diff
@@ -31,7 +31,8 @@ const convertToJsx = (html) => {
             }
 
             const attr = obj.svg[attrKey];
-            const openingTagWithClass = `<svg className={ [this.props.className, '${attr.class}'].join(' ')} `;
+            const dataTag = attr.class === undefined ? "" : `${attr.class.replace("dibsvg-svg ", "")}`;
```
hellatan: can you not get the file name within this script instead?

hellatan: so i think it'd be better to modify the signature to this method when it is called from here: https://github.com/1stdibs/dibs-vg/blob/master/scripts/generateDist.js#L52. There's the `name` variable that you can pass along with the `contents`.

I think it'd be better to pass the `name` argument first then the `contents`:

`htmlToReact(contents)` -> `htmlToReact(name, contents)`

and then in this file, you can use that variable directly instead or do any additional manipulation you need (ie. toLowerCase() maybe)

szhu1026:
@hellatan updated.  could you take a look again?

scripts/generateDist.js
```diff
@@ -49,7 +49,7 @@ svgs.forEach(file => {
 
     // generate react module :
     promises.push(new Promise((resolve, reject) => {
-        htmlToReact(contents).then(result => {
+        htmlToReact(contents, name).then(result => {
```
hellatan: i'd flip flop them: `htmlToReact(name, contents)`

scripts/htmlToReact.js
```diff
@@ -22,7 +22,7 @@ module.exports = Component;
 `;
 };
 
-const convertToJsx = (html) => {
+const convertToJsx = (html, dataTag) => {
```
hellatan: flip flop the signature and then call `dataTag` `fileName` instead: `const convertToJsx = (fileName, html) => { ... }`

szhu1026:
@hellatan made those updates.

hellatan:
@szhu1026 this good to merge now?

szhu1026:
@hellatan yup!

PR #87, opened 2017-06-05
GOODS-6577 Add circle-erase svg

aahill50:
Adding one more svg for reason codes


hellatan:
@aahill50 can you generate the other compiled assets so the icon page can is generated with this new icon?

aahill50:
@hellatan done

hellatan:
⬇️ 〰️ tasks, please @aahill50

aahill50:
@nameofname pointed out last time that I don't have permission to publish to the public repo, so what exactly should I be doing here? Should I still publish to sinopia?

hellatan:
ah damn...forgot about that. But i think you need to publish as the 1stdibs account.

nameofname:
@aahill50 I published minor version 1.39

aahill50:
thanks

PR #88, opened 2017-06-06
flag icon for MCPERSON-4064

orteth01:


sample/compiled/sample.js
```diff
@@ -584,9 +584,9 @@
   \******************/
 /***/ function(module, exports, __webpack_require__) {
 
-	__webpack_require__(/*! /Users/aaron.hill/projects/dibs-vg/sample/js/sample.js */1);
-	__webpack_require__(/*! webpack/hot/dev-server */507);
-	module.exports = __webpack_require__(/*! webpack-dev-server/client?http://localhost:9090/ */509);
+	__webpack_require__(/*! /Users/Teddy/projects/dibs-vg/sample/js/sample.js */1);
```
orteth01: this file was updated when i ran `yarn build`. seems weird to have user-specific paths in a file in our repo.

@hellatan what is this file for? can it be added to .gitignore?

hellatan: this is used for our github page for this repo. The path doesn't really matter since it's commented out. It seems more for reference than anything else. If we were trying to optimize this file, we'd run it through uglify but it doesn't really matter.

orteth01:
@brettjashford looks like i need 1stdibs credentials to publish.
```
npm ERR! you do not have permission to publish "dibs-vg". Are you logged in as the correct user? : dibs-vg
```

do you mind pulling master and publishing for me?

PR #89, opened 2017-06-06
Cleaned up svg's to prevent jest snapshot errors

sjeanat:


sjeanat:
@hellatan I ran it through my fingers if the `<path />` blocks were pretty short. If they were super long I ran it through [here](https://www.gillmeister-software.com/online-tools/text/remove-line-breaks.aspx) to strip out all the white spaces and then I copy-pasted the entire svg in to [here](http://www.freecodeformat.com/svg-editor.php) and made sure it looked like the svgs in the style-guide. Otherwise, it would have taken me sooo long to clean up.

nameofname:
I was literally *just* talking to @robrichard about this - we want to add an SVG minification step to the build script so we don't have to do this kind of work manually. These changes are fine if we visually verify that none of the SVGs are broken, but it would still be better to have this automated.

hellatan:
I think kostas mentioned integrating this package - https://github.com/svg/svgo - to this repo for the automated steps. I think some of the older svgs may have their `fill` attribute set still but I won't hold this PR up b/c of that.

@sjeanat could you maybe publish this as a custom tag and then deploy it to a fruit server so we can take a look in the style guide for any broken svgs?

nameofname:
@hellatan I worked with @sjeanat to build her changes here : 
http://beautifulcode.1stdibs.com/dibs-vg/

All the updated SVGs look good to me.

PR #91, opened 2017-06-23
MCPERSON-3981 Icons for MC

orteth01:
added camera, dot, paperclip, and reply-arrow


orteth01:
@hellatan bump!

src/dot.svg
```diff
@@ -0,0 +1,3 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-dot" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 250 250">
+	<circle cx="125" cy="125" r="125"/>
```
hellatan: why not accomplish this with css?

orteth01: we could certainly do that. we just had the designer make svg versions of all the font-awesome icons we were still using, on of which was `fa-circle`. 

want me to remove this svg?

hellatan: i don't know where this is used so hard for me to fully say if it's just as easy to do it in css vs if you need a svg for this. can you post a screen grab?

orteth01: it is the unread indicator in inbox view

image

not impossible with css but having the icon is certainly easier

hellatan:
can you run all of these svgs through some sort of svg minifier? we shouldn't have line breaks in the `d` attribute within the `path` element. That line break is essentially a space. Take a look at other icons and you'll see what i mean.

We need to add that into our build step so that it cleans it up for you but it obviously isn't there yet.

src/dot.svg
hellatan: i think css would be better b/c if this is in as a svg, other people may use it and it may be useless markup we don't want to serve up on our pages.

orteth01:
> can you run all of these svgs through some sort of svg minifier?

done

hellatan:
published `dibs-vg@1.42.1`

PR #93, opened 2017-06-29
MONEY-5286 add thick star svg

amhorton:
Does this look okay to you?


list.json
```diff
@@ -138,6 +138,7 @@
   "shopping-cart.svg",
   "star-filled.svg",
   "star-outlined.svg",
+  "star-thick.svg",
```
nameofname: @hellatan do we have a naming convention for dibs-vg icons at this point? Should this be called "star-thick-filled"?

hellatan:
what's wrong with this one:

screen shot 2017-06-29 at 5 29 05 pm

hellatan:
@amhorton is there a ticket for this?

amhorton:
@hellatan ticket is here https://1stdibs.atlassian.net/browse/MONEY-5286

That star isn't shaped exactly how design wants, and also looks weird when the outline color is different from the fill color, which is a requirement for this ticket.

hellatan:
> That star isn't shaped exactly how design wants

there should be one star shape, not two. This is ridiculous. I'll comment in jira.

hellatan:
i took a look in .com and brisket helped look throughout the org code for any star references and it looks like it should be safe to update the existing svgs instead of introducing a new one.

amhorton:
I am afraid that even if we swap these svgs out, if we're okay with changing the shape everywhere, that there maybe be styles on the old stars which, when applied to the new stars, don't work properly? Do you think that would be a problem, based on what you guys found? @hellatan @lilianammmatos

lilianammmatos:
I scoured our organization codebase, and I couldn't find any references. I think it's very unlikely that the old star is being used. 

However, if there is an old star somewhere out there, I would rather update it and fix any styling bugs that come up than have two star SVGs.

lilianammmatos:
FWIW the previous star SVG was originally used for .COM Favorites. That icon has since been updated to be a heart.

amhorton:
Ohhhh, it's not being used _at all_, i thought  we were just going to replace the star we had. i will update this pr.

amhorton:
updated!

hellatan:
@amhorton so the current star we have is fine? If so, i don't think you need to delete the outlined version. I really don't think you need to do anything in this package then, right?

amhorton:
@hellatan yeah i'm confused now too, since looking at the diff it looks like i didn't change anything. I probably messed up, let me look into it

amhorton:
Yeah, i didn't copy the file properly `star-filled` is different now.

hellatan:
@amhorton do you need an outline of the star though?

amhorton:
@hellatan no, we don't need it.

hellatan:
i'm not 💯 set with full on deleting the file. Ideally we'd update the outline version to mimic the filled version but since we're not using either one, i'm of two minds with it. @lilianammmatos @nameofname thoughts?

amhorton:
i don't feel strongly about deleting the first, it just seemed weird to keep it if `star-filled` and `star-outline` actually are differently shaped

nameofname:
I think we should delete it. IMO we should do filled icons like we do for the folder icon - eg. use a stroke via CSS instead of an actual filled SVG. It gives a better look to the icons since you don't know what size you will be displaying them at. 

Please be aware that deleting it will necessitate a breaking version change. 

BTW thanks @amhorton for organizing this.

hellatan:
> Please be aware that deleting it will necessitate a breaking version change.

is it breaking if it's not used anymore?!

lilianammmatos:
+1 @nameofname delete it, have only one star and major bump (a minor bump would *probably* be fine because I don't believe this is being used...).

nameofname:
If you guys feel confident that it's not used then let's do a minor change. Updating this in all the repos is a big pain in the ass.

hellatan:
major bump will be a huuuuuuuge pain in the ass.

hellatan:
@nameofname @lilianammmatos - gonna merge this today and publish a minor version

hellatan:
@amhorton please use `dibs-vg@1.43.1` for this one. I brain farted when i was versioning. I'm going to delete both 1.42.3 and 1.43.0 from npm

PR #96, opened 2017-07-17
added CONTRIBUTING.MD derived html to sample file for gh-pages

nameofname:


nameofname:
cc @hellatan as we discussed, I'm including this in the gh-pages documentation so that designers can see it.

hellatan:
not the prettiest output in the actual client (all grey text) but also not a huge priority

PR #99, opened 2017-08-02
TRADE-209: make offer icon

evaldasv:


gedas-monginas:
you have some conflicts, seems like something else got merged and you need to recompile these sample files.

evaldasv:
fixed

gedas-monginas:
published `v1.47.0`

PR #103, opened 2017-08-24
React.PropTypes -> PropTypes

hellatan:


package.json
```diff
@@ -69,5 +69,8 @@
   "peerDependencies": {
     "react": "^15.0.2",
     "react-dom": "^15.0.2"
+  },
+  "dependencies": {
```
nameofname: this should be a dev dep - it's only used in the sample which would get compiled during development.

hellatan: everywhere else has been a regular dep so i've been keeping that pattern.

nameofname: it's not going to be bundled into anything since we always require SVGs out of the dist folder, so *technically* it doesn't matter - but it's so nice to see 0 actual dependencies in your package...

hellatan: FINE RONALD

hellatan:
this can be picked up with whoever bumps it next.

PR #104, opened 2017-08-30
FINDING-11083 Add Affirm payment icon

audriuslubys:


src/payment-affirm.svg
```diff
@@ -0,0 +1,6 @@
+<svg version="1.1" class="dibsvg-svg dibsvg-affirm" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 14.9 9.2">
+    <path fill="#F8F8F8" d="M0 0h14.9v9.3H0z"/>
+    <g fill="#00C8E5">
```
hellatan: I thought the color was supposed to be black? These color should be removed anyhow. I think the first `path` element above shouldn't be apart of this svg as it looks like it should be a background color. If you remove the `g` element, then you should be left with just the second `path` element

hellatan: yup, according to amy's comment, there are two color options for us here. so we'll have to achieve that with css

audriuslubys: This is is for payment options on the pdp, all of those SVGs are in color. This won't be used beneath the price box, that one is displayed by the Affirm api and is displayed as a font instead of a SVG.

gedas-monginas: yes, seems like all `payment-` SVG's have brand colors in them, so makes sense to keep consistent.
The only thing is that this is missing is classes `dibsvg-color-` for potential overwrite's, for example: https://raw.githubusercontent.com/1stdibs/dibs-vg/master/src/payment-discover.svg not sure we really need it, but if talking about consistency, probably we should add them.

hellatan: > that one is displayed by the Affirm api and is displayed as a font instead of an SVG

where is that font coming from?

I would still remove the first `path` element as that only seems like it's supposed to be a background color which we don't need since it is basically white.

audriuslubys: It's loaded and inlined as a base64 by affirm js runtime.

The grey background is the same as other payment icons and I think we should keep it just for the sake of consistency. Adding css classes to override colors might be a solution, as @gedas-monginas suggested.

audriuslubys:
@hellatan does this need updating?

gedas-monginas:
published `dibs-vg@1.48.0`

PR #107, opened 2017-09-15
trash-can: made all 3 same svg, added deprecation warnings

creatyvtype:


scripts/htmlToReact.js
```diff
@@ -22,6 +23,14 @@ module.exports = Component;
 `;
 };
 
+const deprecationWarning = (fileName) => {
+    if (deprecatedFileMap[fileName]) {
+        return `console.warn("dibs-vg | '${fileName}' svg is deprecated. Use '${deprecatedFileMap[fileName]}' <3 Majd");\n`;
```
hellatan: can you use the unicode version of `<3` instead?

list.json
```diff
@@ -155,8 +155,6 @@
   "trade-arrow.svg",
   "trade-logo.svg",
   "trash-can.svg",
-  "trash-filled.svg",
-  "trash-outlined.svg",
```
hellatan: can you do an audit of the entire 1stdibs org to see if these are currently used somewhere?

creatyvtype: My understanding is that this list is only used in the style guides. @nameofname can you verify that? The svgs are being used elsewhere, but now whichever they use will show as the thin outline one, and if it's not referencing "trash-can" will see a deprecation warning. Which reminds me... I need to add a check for env===dev so that we don't see the warnings in production.

nameofname: I'm pretty sure this file isn't used anywhere else. I do recall using it in image upload to require an icon - the old image upload. So if anything check there, but honestly you should just merge this and if a place in the admin breaks where we are using the wrong trash can, then it will force developers to switch and use the right one.

hellatan: there will still be require/import references of `trash-filled` or `trash-outlined` if they are still in use. Use the global search when on github.com/1stdibs or `org:1stdibs` as apart of your search query and you should see something if there are any references

nameofname: @hellatan there definitely still are some uses of those deprecated icons still in use - the point of this is that we don't have to track them all down, they will just become deprecated, and later deleted in a breaking version

creatyvtype: This is just a list of the existing icons. This just removes the icons from that list, but imports will still work of trash-can, trash-filled, and trash-outlined. The latter two will simply display a console.warn error telling it to use trash-can, but the component itself will still render, but all 3 will appear the same way regardless of what's imported.

creatyvtype:
Just want to note that every time I build locally, it updates the sample files with my relative path to the project. I did not commit those changes, but they may be good files to add to .gitignore. Didn't want to do it myself because I did not understand what they were for or what consumes them.

hellatan:
> Just want to note that every time I build locally, it updates the sample files with my relative path to the project

what do those paths look like?

creatyvtype:
@hellatan when i run `yarn build` I get changes to `sample/compiled/sample.js` and `sample/compiled/sample.js.map`. They look like:

```
-/******/ 	var hotCurrentHash = "486e1b88268bc8012daa"; // eslint-disable-line no-unused-vars
+/******/ 	var hotCurrentHash = "92d1eb83f764a4ecec46"; // eslint-disable-line no-unused-vars

...
-	__webpack_require__(/*! /Users/audriuslubys/Projects/dibs-vg/sample/js/sample.js */1);
+	__webpack_require__(/*! /Users/majdmurad/projects/dibs-vg/sample/js/sample.js */1);

...
-	var _UsersAudriuslubysProjectsDibsVgNode_modulesReactTransformHmrLibIndexJs2 = (0, _index6.default)({
-	    filename: '/Users/audriuslubys/Projects/dibs-vg/sample/js/Wrapper.jsx',
+	var _UsersMajdmuradProjectsDibsVgNode_modulesReactTransformHmrLibIndexJs2 = (0, _index6.default)({
+	    filename: '/Users/majdmurad/projects/dibs-vg/sample/js/Wrapper.jsx',

...
-	var _UsersAudriuslubysProjectsDibsVgNode_modulesReactTransformCatchErrorsLibIndexJs2 = (0, _index4.default)({
-	    filename: '/Users/audriuslubys/Projects/dibs-vg/sample/js/Wrapper.jsx',
+	var _UsersMajdmuradProjectsDibsVgNode_modulesReactTransformCatchErrorsLibIndexJs2 = (0, _index4.default)({
+	    filename: '/Users/majdmurad/projects/dibs-vg/sample/js/Wrapper.jsx',
```

hellatan:
oh i wouldn't worry about that file. that's just the source map file for github page so it really doesn't matter.

creatyvtype:
So if I commit it it's ok? It includes the old list.

nameofname:
@creatyvtype it's ok to merge - it's in the pre-publish script, and it needs to get merged periodically so that the gh-pages branch will be updated with the new material.

creatyvtype:
+ dibs-vg@1.49.0

PR #115, opened 2017-10-16
MCPERSON-4590 price-match icon cleanup

nerijussavickas:


hellatan:
I'm not going to publish this version and will be picked up in the next release.

PR #116, opened 2017-10-17
use Node.js trick to properly throw errors from inside .catch!!!!

nameofname:
Node.js swallows errors that are thrown inside catch blocks because catch returns a new promise. This weird trick causes the error to be thrown on the next tick, reference : 

https://stackoverflow.com/a/30741722/4002508

It's important because thrown errors should cause the yarn pre-publish to fail with a non-zero exit code, but instead this causes the process to exit with a code of 0, and a little warning about the unhandled promise rejection. 

That allows the user to publish the package with SVGs MISSING! Very confusing for the developer.


nameofname:
Here's a screen cap of the console output with this fix : 


screen shot 2017-10-17 at 5 03 41 pm

scripts/generateDist.js
```diff
@@ -48,30 +39,29 @@ svgs.forEach(file => {
...
-        function (reason) {
-            checkError(reason);
-        }
-    );
+    .then(() => {
+        console.log("Successfully compiled all SVG files.");
+    })
+    .catch(([err, file]) => {
+        console.log(`Failed to compile SVG file ${file}`);
+        setTimeout(() => { throw err; });
```
robrichard: maybe just `process.exit(1)`?

hellatan: seems like `process.exit(1)` wouldn't give you a great output though compared @nameofname's screen shot

nameofname: @robrichard @hellatan Dale is correct - process.exit(1) causes the compile step to fail, but it's a blunt instrument / lacks the pretty output. I skimmed this article for reference : 
https://www.nczonline.net/blog/2014/02/04/maintainable-node-js-javascript-avoid-process-exit/

nameofname: Rob pointed out that I can get the pretty printed output by using console.error. Small adjustment in this PR : 
https://github.com/1stdibs/dibs-vg/pull/117

PR #131, opened 2018-03-09
DX-2904: new login inquiries/email icon

sashak007:
I ran into this lint warning when committing:
```
dibs-vg/sample/compiled/sample.js
File ignored because of a matching ignore pattern. Use "--no-ignore" to override
```

I didn't pass the `--no-ignore` flag, let me know if I need to explicitly run this file through the pre-commit hooks.


sashak007:
Closing this in favor of the IconWrapper component in dibs-elements.

PR #132, opened 2018-03-12
MCOPOP-5236: Add archive, read, and unread icons

batzlerg:
used in dealer MC as icons for inbox bulk actions


batzlerg:
@brettjashford do you have npm publish rights to this package? everything's merged and versioned appropriately but I don't have permission to publish to the registry

PR #136, opened 2018-04-09
CONT-843 added more-options.svg

Cirakuze:


src/more-options.svg
```diff
@@ -0,0 +1,7 @@
+<svg class="dibsvg-svg dibsvg-more-options" xmlns="http://www.w3.org/2000/svg"
+     xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+     width="250px" height="250px" viewBox="0 0 250 250" style="enable-background:new 0 0 250 250;">
+    <circle class="dibsvg-svg dibsvg-more-options circle-0" style="fill:#6A92BB;" cx="125" cy="62.5" r="41.5"/>
```
moriko: Suggest removing the style with color and only adding the color where the svg icon is used.

PR #139, opened 2018-04-10
Update icon page styling

moriko:
Before | After
------ | ------
1stdibs github io_dibs-vg_ | localhost_9090_


package.json
```diff
@@ -70,5 +70,10 @@
   "peerDependencies": {
     "react": "^15.0.2",
     "react-dom": "^15.0.2"
+  },
+  "dependencies": {
+    "dibs-scss-mixins": "^2.4.3",
```
nameofname: these should be dev dependencies. Not required for consuming modules, only to build the display page.

PR #143, opened 2018-04-25
INTL-486 Add country flags

dericko:
Add us,gb,de,fr flags.

Related ticket: https://1stdibs.atlassian.net/browse/INTL-486


nameofname:
@dericko they get modified when you build, the output is visible here : 
http://1stdibs.github.io/dibs-vg/

When master gets merged to the gh-pages branch it's effectively published.

dericko:
Looks like we can't have `style` attrs

PR #149, opened 2018-05-03
 feat: message-unread-circle svg

PeteJodo:


src/message-unread-circle.svg
```diff
@@ -0,0 +1,4 @@
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 250 250">
```
nameofname: you need to add the class name to this svg. See any other svg for an example, there should be 2 classes, one for dibs vg and another one including the file name.

moriko:
Wait what? We do have similar icon:
screenshot 2018-05-04 09 26 28
Why do designers keep creating alike icons? We should try do help the designers by showing existing icons in http://1stdibs.github.io/dibs-vg/ so that svg icons wouldn't be cluttered with unused or inconsistent icons.

PeteJodo:
That's an incorrect icon @moriko. They want to be able to color the circle independently from the envelope. I would have removed it but it's currently being used in other spots and it isn't a 1-to-1 replacement with what I put in here. I'll look into removing the old one when it's been updated

PR #151, opened 2018-05-18
Update IE11-specific gotchas in README

PeteJodo:


README.md
```diff
@@ -135,4 +135,6 @@ Inline SVGs are great, not only because they're more performant, but they're als
 
 ## Gotchas :
 
-A SVG will not be compiled into the `dist` folder properly if the opening `<svg>` tag contains the `xml:space="preserve"` attribute.
+- A SVG will not be compiled into the `dist` folder properly if the opening `<svg>` tag contains the `xml:space="preserve"` attribute.
+- The SVGs automatically have styling that adds height of `100%` and width of `100%`. In non-IE11 browsers, setting just one of those to an explicit value (e.g. `height: 25px`) correctly sizes the svg container element to the correct size but in IE11 it adds whitespace in the dimension that wasn't explicitly set. Make sure to add sizes to both dimension (e.g. `height: 25px; width: 25px;`)
```
hellatan: reword this part for clarity:
```
The SVGs automatically have styling that adds `height: 100%` and `width: 100%`. In non-IE11 browsers, setting just one of those to an explicit value (e.g. `height: 25px`) on the containing element correctly sizes the svg container. However, IE11 adds whitespace in the dimension that was not explicitly set. To fix this, make sure to add sizes to both dimension (e.g. `height: 25px; width: 25px;`).
```

README.md
```diff
@@ -135,4 +135,6 @@ Inline SVGs are great, not only because they're more performant, but they're als
...
+- The SVGs automatically have the css rule `fill: currentColor;` (which I couldn't find any reference to on MDN!) which is supposed to set the fill of an SVG with an inherited `color` rule. This doesn't work in IE11. Use `fill: <color>`.
```
moriko: How `currentColor` is not supported in IE11? https://caniuse.com/#feat=currentcolor ??

PeteJodo: you're right actually. The issue I'm talking about is actually caused by something different where the `currentColor` doesn't applied on rules that are explicitly set for children of the element that has `currentColor` set:

capture2_original

The black circle should be red and that's because the color is applied directly on the circle which is a child of what has `currentColor` set so it doesn't apply. I'll removed this

PR #154, opened 2018-06-07
Newer better zoom icons

cheryly279:


src/zoom-in.svg
```diff
@@ -1,4 +1,7 @@
...
-    <path d="M283.232,182.728h-63.954v-63.953c0-2.475-0.905-4.615-2.712-6.424c-1.809-1.809-3.951-2.712-6.423-2.712H191.87c-2.474,0-4.615,0.903-6.423,2.712c-1.807,1.809-2.712,3.949-2.712,6.424v63.953h-63.954c-2.474,0-4.615,0.905-6.423,2.712c-1.809,1.809-2.712,3.949-2.712,6.424v18.271c0,2.475,0.903,4.617,2.712,6.424c1.809,1.809,3.946,2.713,6.423,2.713h63.954v63.954c0,2.478,0.905,4.616,2.712,6.427c1.809,1.804,3.949,2.707,6.423,2.707h18.272c2.473,0,4.615-0.903,6.423-2.707c1.807-1.811,2.712-3.949,2.712-6.427v-63.954h63.954c2.478,0,4.612-0.905,6.427-2.713c1.804-1.807,2.703-3.949,2.703-6.424v-18.271c0-2.475-0.899-4.615-2.703-6.424C287.851,183.633,285.709,182.728,283.232,182.728z"/>
+<svg version="1.1" class="dibsvg-svg dibsvg-zoom-in" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 250 250">
+	<polygon points="97.7,59.7 87.7,59.7 87.7,90.7 57.8,90.7 57.8,100.7 87.7,100.7 87.7,129.5 97.7,129.5 97.7,100.7 127.8,100.7
```
hellatan: can you please make sure to put all attribute values on the same line? newlines = single space on the same line

src/zoom-out.svg
```diff
@@ -1,4 +1,6 @@
-<svg version="1.1" class="dibsvg-svg dibsvg-time" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 475.084 475.084">
```
hellatan: it's weird that `dibsvg-time` was the class name for this to begin with, but can you do a search for that class throughout the 1stdibs org? If there are references to this, this would technically be a breaking change, but i suspect that there will be no references to it b/c it is obviously the wrong name for this svg.

cheryly279: Yes it is super weird that the class name was dibsvg-time, and yes I checked it's not being used anywhere.

src/zoom-in.svg
cheryly279: newlines removed

PR #155, opened 2018-06-11
fix: added comment for using default registry url

hellatan:


.npmrc
```diff
@@ -1 +1,7 @@
+# being behind a proxy has potentially for appending
```
nameofname: This sentence is grammatically off, but overall the comment makes sense to me

hellatan: this is why i'm a developer, not an english teacher 😄

hellatan: oh but this is bad...i'm going to fix it hahaha

hellatan:
not publishing as this will get picked up in the next bump

PR #156, opened 2018-06-11
Add svgo

hellatan:
the first commit is where the actually changes reside. the second commit is just formatting svgs with the updated space indentations.


.svgo-config.json
```diff
@@ -0,0 +1,49 @@
+{
+  "plugins":[
+    {"cleanupAttrs": true},
```
hellatan: i only wanted this feature so i only enabled this plugin. These are all the current plugins and we can enable them as we see fit.

package.json
```diff
@@ -9,15 +9,19 @@
...
     "postversion": "git push upstream master && git push --tags"
   },
-  "pre-commit": {
-    "run": [
-      "gitlint"
+  "lint-staged": {
+    "*.svg": [
+      "yarn format:svg"
+    ],
+    "*.js": [
```
hellatan: this was the `gitlint` script but i removed `pre-commit` in favor of `husky` and `lint-staged`

src/minus.svg
```diff
@@ -1,3 +1,3 @@
 <svg version="1.1" class="dibsvg-svg dibsvg-minus" viewBox="0 -120 250 250" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-    <rect width="250" height="15" rx="10"></rect>
-</svg>
\ No newline at end of file
+    <rect width="250" height="15" rx="10"/>
```
hellatan: 🤔 it does seem to self-close redundant tags though

hellatan:
@nameofname @moriko thoughts?

moriko:
I'm more than happy that the attributes would be cleaned up.
Have you tried out the `addAttributesToSVGElement`? As we have few attributes that should be added on svg this would be great if was enabled at some point.

hellatan:
> Have you tried out the addAttributesToSVGElement?

@moriko i did not. I'll look into that today and report back

src/app-store.svg
```diff
@@ -1,35 +1,33 @@
 <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" class="dibsvg-svg dibsvg-app-store" version="1.1" viewBox="0 -47 135 135">
-    <g>
```
hellatan: these SVGs look the same even with the removed group tag

.svgo-config.yml
```diff
@@ -0,0 +1,53 @@
+quiet: true
```
hellatan: it seems like the preferred config is a yaml file

hellatan: @moriko i also did not look at all the plugins here, so please suggest any other ones we may want to use. These are the only ones that are currently on:
- cleanupAttrs
- addAttributesToSVGElement
- sortAttrs

src/app-store.svg
moriko: Grouping just add elements in folder in illustrator. If there's no additional attribute then it's save to remove.

hellatan:
@nameofname @moriko ok, this is ready to be merged. I'm waiting for svgo to be published so that the `xmlns` attribute isn't always the first one on the `svg` node, but that can be updated once it has been published (https://github.com/svg/svgo/pull/984)

package.json
```diff
@@ -50,22 +55,31 @@
     "dibs-scss-mixins": "^2.4.3",
     "dibs-scss-vars": "^3.6.4",
     "dibs-styles": "^2.4.0",
-    "dibslint": "^2.0.3",
```
nameofname: 👍

hellatan: linting #rocksolid

hellatan:
`dibs-vg@1.73.0`

src/logo.svg
```diff
@@ -1,10 +1,10 @@
-<svg version="1.1" class="dibsvg-svg dibsvg-logo" title="1stdibs.com" viewBox="543 138 2021.4 500" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
gedas-monginas: @hellatan do you think loosing a `title` from 1stdibs logo might be an issue? SEO? :)

hellatan: hmm...not sure tbh. i'll see to try and add it back.

hellatan: actually, it seem like there is no `title` attribute or svgs. a `<title>` element would have to be added to this svg instead so it seems like this attribute has been doing nothing for us. I'll add the `<title>` element though

hellatan: @gedas-monginas - https://github.com/1stdibs/dibs-vg/pull/158

PR #158, opened 2018-06-22
fix: format svgs, added <title> for logo

hellatan:


gedas-monginas:
so if `title` attribute wasn't doing anything maybe we don't need it at all? :)

hellatan:
i'm ok with leaving it out 😄

hellatan:
the change is in here. I'm going to merge as-is.

hellatan:
these changes can get picked up with whoever bumps this package next

PR #159, opened 2018-07-05
feat: GROWTH-4128 Sold icons

martynas-b:


moriko:
@martynas-b you really don't need to have text svg that is text. First is won't be intl'ed, second you can just use text there as I saw from the designs.

And also `dlos.svg`? That's in revers `sold` :D What is going on?

list.json
```diff
@@ -48,6 +49,7 @@
   "credit-card.svg",
   "customer-support.svg",
   "diamond.svg",
+  "dlos.svg",
```
martynas-b: That is actually a word "Sold" icon, but we want to hide it from google.. :)

martynas-b:
@moriko - that is exactly what we want to do :) no word "sold"! neither as html nor svg text. as well as icon name :) yes, i know it looks strange, but thats exactly what we are doing.

moriko:
First. Google doubtfully will see the name of the svg file? Am I wrong? For example "nay-pdp-en" for the class on the svg would be fine as all the text should be translatable even if it's a svg and if that's not so internationally used word.

p.s. still against having text as svg in the `dibs-vg`. Would probably be better to have component that gets locale and outputs the correct texted svg.

martynas-b:
It was agreed not to use any locales on this. it's gonna be the same for all. currently it's gonna be as a test to find out if that helps to trick the google. icon name itself is not visible, but attribute class="dibsvg-svg dibsvg-dlos" will be.

gedas-monginas:
@hellatan it will be used in ABF for SOLD PDP's, will be running SEO A/B. Instead of text SOLD we will show icon which says `SOLD` or that other icon 🚫 
We can definitely put it straight on the page until we have a winner and one of the variants is 💯

martynas-b:
I'm closing this PR for now. Will add the svgs to the component itself. When there are test results we could add needed svg to dibs-vg then.

hellatan:
Great thanks @martynas-b

PR #163, opened 2018-07-24
feat: FINDING-12618 Update zoom in-out

moriko:
Small step for an icon but giant leap for UX.

artboard 2


moriko:
`1.74.2`

src/zoom-in.svg
```diff
@@ -1,4 +1,5 @@
 <svg xmlns="http://www.w3.org/2000/svg" class="dibsvg-svg dibsvg-zoom-in" viewBox="0 0 250 250">
+    <path d="M93.3,9 C46.8,9 9,46.7 9,93.1 C9,139.4 46.8,177.2 93.3,177.2 C139.8,177.2 177.6,139.4 177.6,93.1 C177.6,46.7 139.8,9 93.3,9 Z" fill-opacity="0.3" fill="#FFFFFF"></path>
```
hellatan: 🤔...it seems like the formatting didn't run on these svgs b/c i just pulled down and ran `yarn format:svg` and both of these svgs updated which they shouldn't have if the formatter ran at the right time. I just don't think the scripts are setup properly b/c i bet, @moriko, if you go to your local, you'll see these two files changed but not committed.

moriko: Weird :/ I ran the `yarn format:svg` before but only now got the changed svgs :/

hellatan: weird...it should have run when as a precommit hook

PR #164, opened 2018-07-24
feat: TRADE-2603 added icon for trade sourcing

hellatan:
thoughts on a different name for this icon? Open to suggestions...

screen shot 2018-07-24 at 4 24 57 pm


moriko:
`chair-search`? 😆
`chair` -  well as the it's the main object in the svg
`search` - as this magnifying glass is used for the search icon 😏

hellatan:
What about just `sourcing`? That's what the context of the icon is. Do we want to be super literal with this?

hellatan:
@moriko i spoke to design and they said `sourcing` was fine since this is what the icon is going to be initially used. I think in general this is broad enough of a name as well rather than getting too literal with it.

hellatan:
also, `sourcing` is the equivalent of `search` just to be clear

PR #166, opened 2018-07-26
fix: renamed icon to be more literal

hellatan:
i'm going to unpublish 1.75.0 and 1.75.1


hellatan:
cc @evaldasv

hellatan:
dibs-vg@1.75.2 published

PR #169, opened 2018-08-27
fix: remove extra space from checklist svg

PeteJodo:


src/checklist.svg
```diff
@@ -1,5 +1,5 @@
...
-    <path d="M186.3,101.8L128,152.4l-20.7-27.9c-1.5-2-4.3-2.4-6.3-0.9c-2,1.5-2.4,4.3-0.9,6.3l23.6,31.8c0.7,1,1.9,1.7,3.1,1.8 c0.2,0,0.3,0,0.5,0c1.1,0,2.1-0.4,2.9-1.1l62-53.8c1.9-1.6,2.1-4.5,0.5-6.3C191,100.4,188.2,100.2,186.3,101.8z"/>
-    <path d="M171.9,136c0.1,1.4,0.2,2.7,0.2,4.1c0.1,26.7-21.6,48.5-48.3,48.5s-48.5-21.6-48.5-48.3c-0.1-26.7,21.6-48.5,48.3-48.5 c16.7-0.1,30.4,7,39.2,19.8l6.7-6c-10.4-14.7-26.5-22.8-45.9-22.8c-31.7,0.1-57.3,25.9-57.2,57.6c0.1,31.7,25.9,57.3,57.6,57.2 c31.7-0.1,57.3-25.9,57.2-57.6c0-3.6-0.4-7.1-1-10.5L171.9,136z"/>
+<svg xmlns="http://www.w3.org/2000/svg" class="dibsvg-svg dibsvg-checklist" viewBox="0 0 188.1 250">
```
moriko: Interesting :) removing extra space but the svg isn't square anymore and now visually has more space :)

PeteJodo: yeah I think adding empty space just to have svgs be squares is flawed. It makes implementing designs where alignment matters very hacky

moriko: Think that in a way the design should be made so that we wouldn't have to hack things. Also what if this icon in different place will have different alignment then this viewBox change will be a hindrance one way or the other.

PeteJodo: it's only in 2 locations currently both of which won't be effected.

> Think that in a way the design should be made so that we wouldn't have to hack things. 

that would be to have svgs as their true size rather than having arbitrarily empty space just to have them be an arbitrary square. The other option would be to have icons all be perfect squares

> Also what if this icon in different place will have different alignment then this viewBox change will be a hindrance one way or the other.

Not necessarily true unless someone already did a hacky way to make that alignment and in which case, it should be fixed anyway (confirmed that isn't the case though). Vertical alignment won't change since it's still 250 tall, which is the most common form of alignment. Currently horizontal alignment is effectively impossible and design should not at all have to design around no horizontal alignment with svgs, that's just cray 😏

hellatan: usually we have the svgs to their true size with no whitespace around them. If they have whitespace around them, then it's due to the fact that it's hard to tell from a PR what the spacing should be like. If you look at any of the 1stdibs logos that are irregular sizes, you'll see that the viewbox values are indeed the size of the svgs without any whitespace around them.

PeteJodo: here's an example of a layout where this is an issue

tradeonboardpdpmodal

PR #170, opened 2018-08-29
feat: GROWTH-4275 Sold icon

martynas-b:


src/sold.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" class="dibsvg-svg dibsvg-sold" viewBox="0 90 250 70">
```
martynas-b: Google shouldn't scan class names I believe so decided to leave it 'sold'.

hellatan: Why is the viewbox `0 90...` shouldn't those two numbers be `0 0`?

hellatan: also, what about translations?

martynas-b: This word 'sold' is actually vertically placed in the middle of a square icon, but as we are trying to pretend it's a real text and put it inline with other normal texts it can't be a square and should only view boxed as a rectangular around this 'sold' text. so we can put it inline, match sizes and don't mess line heights.

martynas-b: sure, will do it. at some point it was in parent ticket but I guess it just got lost for the test removal tickets. will have a look tomorrow.

PR #174, opened 2018-09-24
feat: TRADE-2933 make buyer-assurance icon more customizable

PeteJodo:
sure to be a contentious PR. Added the ability to "fill-in/invert" the `buyer-assurance` icon as well as gradient. The idea is this is a stopgap to get it in now until we add the ability to add gradients in svgs. I didn't create the ticket yet but it will be a FI ticket and will link to it when I create it


PeteJodo:
FI ticket to remove this later https://1stdibs.atlassian.net/browse/FI-628

src/buyer-assurance.svg
```diff
@@ -1,4 +1,11 @@
...
+<svg xmlns="http://www.w3.org/2000/svg" class="dibsvg-svg dibsvg-buyer-assurance" viewBox="0 0 250 250">
```
hellatan: should this viewBox value have changed so drastically? does it really matter that much since the aspect ratio is the same?

PeteJodo: yeah so I thought this was weird too but what actually happened was that it originally was done on a 518x518 view box, Jen fixed it to 250x250 and since the points within the svg are with respect to view box size I couldn't keep it the same as it already was (otherwise the design itself only filled about half the view box)

PeteJodo:
@hellatan yes I tested it, since the svg is just some html I was able to paste it in directly and have it work just fine

PeteJodo:
@hellatan bump

hellatan:
this can be a patch bump @PeteJodo

hellatan:
published `dibs-vg@1.77.1`

PR #175, opened 2018-09-25
fix: add Jenkins build support

hellatan:


hellatan:
i added this to the jenkins job that we are going to use for publishing

.npmrc
```diff
@@ -1,3 +1,8 @@
+# if you are publishing locally, you must add
+# `export DIBS_NPM_TOKEN="<token value>"`
+# to your `~/.profile` or equivalent file
+# otherwise publish from the jenkins job
+//registry.npmjs.org/:_authToken=$DIBS_NPM_TOKEN
```
hellatan: `$DIBS_NPM_TOKEN` seems to only work in zbash shells but will work in bash shells as well.

hellatan:
>  i was thinking we could add a script (maybe in cutman) that would be able to clone and publish any of our packages. Instead of having to manually add a new script and job to each one.

@robrichard I think we would only need to do that for the (short) list of public packages we have, right? Since we're using lerna for our private packages to already auto publish.

.npmrc
twhid: This comment is very confusing to me.

hellatan:
oh, just thought about this - maybe this gets kicked off whenever something is merged to master so then it can post back which version was published.

.npmrc
hellatan: huh...me too... 🤷‍♂️

hellatan: @twhid that make more sense?

twhid: sorry i meant the PR comment:
> $DIBS_NPM_TOKEN seems to only work in zbash shells but will work in bash

Does it ONLY work in zbash or does it work in zbash and bash?

hellatan: oh ya, i realized that after I fixed my confusing shell comment.

> Does it ONLY work in zbash or does it work in zbash and bash?

it works in both - I tried this locally and it did not complain. I built it on jenkins and jenkins did not complain.

hellatan:
authorization changed @moriko. I haven't had a chance to work on this further b/c of more pressing changes that need to get deployed sooner rather than later. In the mean time, i'm ok with publishing manually so feel free to ping me 😄

hellatan:
@moriko yes, closing.

PR #176, opened 2018-09-26
fix: convert fill-opacity and stop-color to camelCase

PeteJodo:


hellatan:
published `dibs-vg@1.77.2`

PR #179, opened 2018-10-05
feat: FINDING-13172 SVG optimization

moriko:


moriko:
Also tried the `mergePaths` but as wasn't in love what it have done to visa icon decided to not use that at the moment. Unless we would create list for icons that should not be svgo'd 🤷‍♀️ 

The visa `(r)` blobed 😒 but maybe it's not a big deal but wouldn't want branded icons change their look in any way.

before | after
------ | ------
screenshot 2018-10-11 15 11 58 | screenshot 2018-10-11 15 10 07

hellatan:
@moriko i think what you have now is fine

hellatan:
@moriko @nameofname @creatyvtype i'm going to make this a patch bump since it's really just a clean up task. Thoughts?

moriko:
Yeah this should go out as a patch 🙇‍♀️

hellatan:
published `dibs-vg@1.78.1`

PR #180, opened 2018-10-16
feat: TRADE-2882 professional-wreath graphic

nivers:
https://1stdibs.atlassian.net/browse/TRADE-2882


src/wreath-professional.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" class="dibsvg-svg dibsvg-wreath-professional" viewBox="0 0 250 250">
```
hellatan: @nivers i didn't realize that your question about this transparency was for an icon going into dibs-vg - i thought it was for the more trade-specific svgs you needed to clean up. You should add a `class` on the `<path>` element instead and then target that class in the implementation instead since we don't provide any sort of stylesheet with these svgs.

hellatan: we can leave it

PR #181, opened 2018-10-18
feat: compile style attributes

nivers:
Not pretty but it's a proof of concept. I'm not sure `xml2js.Builder` provides a way to do this in a cleaner way


scripts/htmlToReact.js
```diff
@@ -34,6 +35,21 @@ const deprecationWarning = (fileName) => {
     return "";
 };
 
+const convertStyleStringToObjectString = node => {
+    const attributes = node[attrKey];
+    const styleAttr = attributes[styleKey];
+    if (styleAttr && typeof styleAttr === 'string') {
+        const styleStrings = styleAttr.split(';').filter(Boolean);
+        const keysAndValues= styleStrings
```
hellatan: i guess there's not linter here - missing a space before the `=`

scripts/htmlToReact.js
```diff
@@ -34,6 +35,21 @@ const deprecationWarning = (fileName) => {
...
+            .map(styleString => {
+                const [key, value] = styleString.split(':');
+                return `${key}: ${value}`;
```
hellatan: will the `value` have quotes around it if that's how the `style` attribute has the property value? For example, `opacity: 0.25` is fine without quotes. However, and I'm not saying this will happen, but if someone does `font-family: "Helvetica, sans-serif"` - will the quotes around `"Helvetica, sans-serif"` be retained? 

Inadvertently, I just realized that you need to camelCase any `key` that is hyphen-ated to match JS syntax.

scripts/htmlToReact.js
```diff
@@ -48,8 +75,18 @@ const convertToJsx = (fileName, html) => {
...
+            // this will need to be made recursive if styles are ever added to children's children
+            const children = obj.svg.path;
+            if (Array.isArray(children)) {
+                children.forEach(child => convertStyleStringToObjectString(child));
+            }
+
             const xml = builder.buildObject(obj);
-            const xmlString = xml.replace('<svg ', openingTagWithClass);
+            const xmlString = xml
+                .replace('"{{', '{{').replace('}}"', '}}') // remove quotes surrounding style objects
```
hellatan: can you add an example to your comment of before and after? It should be something like this, right:
```
// style="{{ opacity: .25 }}" => style={{ opacity: .25 }}
```

list.json
```diff
@@ -214,4 +214,4 @@
   "yoga.svg",
   "zoom-in.svg",
   "zoom-out.svg"
-]
+]
```
hellatan: i don't think you need this change, please revert this.

hellatan:
published `dibs-vg@1.79.1`

PR #184, opened 2018-10-23
feat MODIFY-SVG-CONFIG: modify .svgo.config.yml build settings

liormb:


.svgo-config.yml
```diff
@@ -11,10 +11,10 @@ plugins:
   - inlineStyles: false
   - minifyStyles: false
   - convertStyleToAttrs: false
-  - cleanupIDs: true
+  - cleanupIDs: false
```
moriko: Why would you need id?

liormb: ping @dale

hellatan: @moriko - https://github.com/1stdibs/dibs-vg/blob/master/src/buyer-assurance.svg - there's a `id` in there for a gradient:
```
    <defs>
        <linearGradient gradientTransform="rotate(90)" id="dibsvg-buyer-assurance-buyer-gradient">
            <stop offset="0" stop-color="#F7F4EB"></stop>
            <stop offset="1" stop-color="#C2A661"></stop>
        </linearGradient>
    </defs>
```
https://raw.githubusercontent.com/1stdibs/dibs-vg/master/src/buyer-assurance.svg

and @PeteJodo tried using a `class` instead when he was applying the gradient, but it would only work with an ID

PeteJodo: specifically, it's the `url()` css function that doesn't support class names since to apply the gradient you have to do `url(#dibsvg-buyer-assurance-buyer-gradient)`

hellatan: if we start to do more gradients within svgs, then we'll unfortunately have to keep this. We could potentially submit a PR to svgo to modify this flag with a whitelist of elements to ignore too. I did that for one of the config options.

PR #187, opened 2018-10-30
fix: TRADE-3271 fix buyer protection icon

ex13:


hellatan:
published `dibs-vg@1.81.1` - @ex13 please update dbl with this version and the downstream bump job. I thought I had those docs outlined somewhere but I can't find them. I think @nitruxa or @evaldasv should be able to help you out with that update.

PR #191, opened 2018-11-19
feat: TRADE-3401 udpdate message icon

jdk243:
https://1stdibs.atlassian.net/browse/TRADE-3401


hellatan:
@jdk243 can you run this through `yarn format:svg` and see if there are any diffs you need to commit? I thought it was going to fail the pre-commit hook, but it just does it for you and then any updated svg does not get added, actually. Something i need to look into.

jdk243:
@hellatan will do

hellatan:
published `dibs-vg@1.83.1`

PR #192, opened 2018-11-20
feat: FTV-121 Negotiable icon

martynas-b:
Relates to: https://github.com/1stdibs/app-buyer-finding/pull/5031


hellatan:
@martynas-b - we just updated the current `message.svg` icon yesterday - https://github.com/jdk243/dibs-vg/blob/d1e5ee9078c1135d008303d8063c2060d6227f35/src/message.svg - i think the message bubbles should have rounded corners.

Also, i need to fix this, but can you run `yarn format:svg` before committing? It does it as a pre-commit hook but doesn't fail if there are any changes. Something I need to look into fixing.

martynas-b:
@hellatan - I see, I didn't notice that message icon was changed. this was the icon for this test, but I assume we should keep it consistent with message icon. I'll try to change it tomorrow. by the way, I think this PR svg was already formatted. I noticed that it made changes in pre-commit and I commited them again.

hellatan:
@martynas-b sounds good. we literally just changed it yesterday but the gh page hadn't been updated yet. I just merged that all in today so it should be the most up to date now.

hellatan:
published `dibs-vg@1.84.0`

